PR #10, opened 2016-11-08
Changes I made while setting the acquisition import and the montage services

cgoina:
Eric,

From my local repo I removed the intellij files because they conflicted with mine and instead of checking mine in I chose to remove them from the repo. Also I made changes to the pom.xml to use 1.8 as source and target because I used lambdas.


render-ws-java-client/src/main/resources/logback.xml
```diff
@@ -7,9 +7,62 @@
...
         </encoder>
     </appender>
-  
-  <root level="debug">
-    <appender-ref ref="STDOUT" />
-  </root>
 
-</configuration>
\ No newline at end of file
+    <appender name="FILE-INFO" class="ch.qos.logback.core.rolling.RollingFileAppender">
```
cgoina: This is the log config that I used for the service setup. The idea was to be able to use different log files for info messages and for error messages.

trautmane:
I merged these changes with this commit: [1dabb5a](https://github.com/saalfeldlab/render/commit/1dabb5a2d8eff85e36acb7967bbb48e7ef2fcd09)

Issue #4, opened 2016-10-19
Override tilespec attributes at render time (specifically, minIntensity and maxIntensity)


perlman:
A "tilespec override" mechanism would be nice for making one-off visualization-only changes to data in the database.

The specific use case we have is to play with the minIntensity & maxIntensity values for data where they were incorrectly specified.

fcollman:
I believe we have implemented minIntensity and maxIntensity in the latest changes that eric merged from my branch, so this should be marked as complete.

Issue #1, opened 2015-01-08
RenderParameters and TileSpec should extend a common base interface


axtimwalde:
RenderParameters and TileSpecs should both be possible sources to Rendering as Rendering generates an image.
In particular, we see it replicating bounding box estimation, and meshCellSize.  There are two bounding boxes in RenderParameters at this time, the minimal bounding box that is derived from the minimal bounding boxes of all included TileSpecs and the bounding box of the rendering `canvas' (x,y, width, height).  We currently believe that there should be only one box.  In RenderParameters and TileSpec, this box can be estimated by deriveMinimalBoundingBox(...), RenderParameters additionally adds padding and uses meshCellSize from internal state whereas TileSpec does not offer padding and gets meshCellSize as a parameter.  Unifying this seems necessary.

PR #16, opened 2017-01-18
get webUI name dynamically

fcollman:
Here is a pull request that executes one of the options.. to add the manage-stack.sh shell script to the render path.... other option is to change the EM aligner code to point to manage_stacks.sh script that exists here.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/RenderParameters.java
+++ b/render-app/src/main/java/org/janelia/alignment/RenderParameters.java
@@ -91,11 +91,11 @@ public class RenderParameters implements Serializable {
     @Parameter(names = "--area_offset", description = "add bounding box offset", required = false)
     public boolean areaOffset;
 
-    @Parameter(names = "--minVal", description = "minimum intensity value to set for all tilespecs", required = false)
-    public double minVal;
+    @Parameter(names = "--minIntensity", description = "minimum intensity value to set for all tilespecs", required = false)
+    public double minIntensity;
 
-    @Parameter(names = "--maxVal", description = "maximum intensity value to set for all tilespecs", required = false)
-    public double maxVal;
+    @Parameter(names = "--maxIntensity", description = "maximum intensity value to set for all tilespecs", required = false)
+    public double maxIntensity;
 
     @Parameter(names = "--gray", description = "convert output to gray scale image", required = false)
     public boolean convertToGray;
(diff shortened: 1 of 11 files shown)


fcollman:
I've modified this to remove the manage-stack.sh add, and added the functionality to use the dynamic grabbing of the webUI... however, this does depend on upgrading to spark 2.0, though it passed all the tests on my machine.  I'll leave it here for you to look at. @trautmane

fcollman:
these extra commits now implement the feature outlined in 
https://github.com/saalfeldlab/render/issues/4
at least for the box/ /jpeg-image call, which is what currently used by ndviz and other dynamic rendering services.   Could easily be added to further things.  Also fixed a current bug with the download link for SLF4J which needed to be https

trautmane:
Thanks for working on this Forrest! 
Sorry I was not able to look at your changes until now.

Here are my comments on the different "development threads" in this request:

1. Commit: Change manage_stacks.sh to manage_stack.sh so that older EM_aligner tool integration works.  

My comments: I don't think this is needed anymore, since Khaled has updated the EM_aligner code base to reference manage_stacks.sh.  Let me know if we're still missing something.

2. Commit: Change executor logging utility to work with Spark 2.0.  

My comments: I'm glad you found this solution.  Unfortunately, we have to work with Spark 1.6.2 at Janelia for a while (not sure how long).  Since this is not critical functionality (just some extra log info) what do you think about just commenting out the long term solution so that it is backward compatible for now?

3. Commit: fixed download link.  

My comments: thanks!

4. Commit: added minVal and maxVal render parameters.  

My comments: Nice feature!  I prefer explicitly naming the parameters minIntensity and maxIntensity instead.  What do you think?   

@fcollman

fcollman:
Hi Eric,

#1 I agree... that's why I removed it in the later commit
#2 I understand... I'll comment it out for others to use if they want.. if i'm feeling ambitious i'll try implementing the more hacky version of look at 4040 and then look up to 4060 and then give up.  Should only look at 4040 most of the time so shouldn't be too slow, but will fallback gracefully, and be backwards compatible with 1.6.  Mostly I want to stay close to your master as possible.
#3 you are welcome!
#4 I am fine with renaming these to minIntensity and maxIntensity.

My only question is now on github etiquette... should i withdraw this pull request and submit a different one? should i do so after bringing my branch back to master, or just tag on extra commits that bring the final code into a place we agree upon?  I'm a newbie at this kind of thing.

fcollman:
Hi Eric, I have finished implementing the changes you requested.  I have also expanded the point match services deletion methods to allow for cleaning up mistakes in point matches, mostly refactoring code you already wrote in different combinations.  I also did implement the hacky version of scanning ports for the proper appId, and tested it on my system.. you should probably test it at janelia as well though as we have different versions of spark.

PR #29, opened 2017-03-23
Remove all transforms on matching

fcollman:
adding an option to allow different behavior for normalizeForMatching that simply removes all transformations, better for truly "local" coordinates and longer lists of transformations.


trautmane:
Hi Forrest - Are you using this?  I just wanted to make sure before I looked more closely at it.  Thanks, Eric

fcollman:
I am using this in order to allow us to remove all the transforms when doing pointmatching rather than having to stick with the rigid definition of transformations that the solver expects.... so we need this, or need a more flexible way of tagging "local" transformations vs "global" transformations, and having normalizeformatching remove all non "local" transformations.

fcollman:
I'm closing this in favor of the recent commits

PR #37, opened 2017-06-13
Minmax filter fix

fcollman:
Stephan came to visit and we found out a problem with the way that min and max intensities and filters interact, because the min  and max intensities were not being set before the filter was applied.  Because files can set their own min and max intensities in the image processor, this was causing problems with some tif files.  Applying the min and max intensities before the filters makes more sense.


fcollman:
I may have gotten my branches screwed up.. but this is the important commit.
https://github.com/saalfeldlab/render/pull/37/commits/cc63df9fd6034d88574f6538d8348b6269c839a8

trautmane:
Thanks Forrest.  I need to figure out how this fits into the new structure I just merged into the master branch.  I'll let you know when I've sorted through it all.

trautmane:
To clean up this pull request, I accepted the changes but essentially undid them.  I then applied what I think are the correct corresponding v2 changes with this commit: [8c7f89f](https://github.com/saalfeldlab/render/commit/8c7f89f29393786990f3e9c1b5e2d3ef91a076bf) .  I may need to review this with Stephan.

fcollman:
Awesome I can confirm this worked

PR #39, opened 2017-06-14
Dockerfile java install fix

fcollman:
This is a fix for the docker install of render which install java8 from oracle via ubuntu package management, rather than the link which is presently broken... hope this will be more robust.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,7 +60,6 @@ ENV JETTY_RUN $JETTY_BASE/logs
 ENV JETTY_PID $JETTY_RUN/jetty.pid
 ENV JETTY_STATE $JETTY_RUN/jetty.state
 
-ENV JAVA_HOME /var/www/render/deploy/jdk1.8.0_73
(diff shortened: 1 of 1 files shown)


trautmane:
I think the best long-term Docker solution is to start with a base Jetty image (instead of Ubuntu), deploy an already built render war, and skip the Git clone, JDK install, Maven build, etc. steps.  I've played with this a little and ultimately hope to make it available on DockerHub as a render image which can simply be installed - ETA July?

fcollman:
agree 100%

fcollman:
actually... what do you mean by already built render war?  I'm not sure I
agree with that, I think the dockerfile should explicitly show you how to
compile the war file, so that others can change the code and then change
the dockerfile if they want.  You can still make the docker image available
on DockerHub with the war already inside, so people can just download and
run it easily, and still modify the configs if they want by extending the
Dockerfile, but having the compile process documented and mechanized is
very important.

trautmane:
Agreed - supporting both cases (just-use and modify+build+use) is important.

PR #43, opened 2017-09-12
S3 URL support in render

perlman:
This is the current version of my code for reading data from AWS s3.

This consists of creating a AWSCredentialsProvider for handling credentials (via environmental variable, .aws/config or instance IAM role) and an extension of ij.io.Opener for catching the S3 URLs.

Suggestions welcome.  I am not a big fan of how I intercept the calls to Opener.


trautmane:
Sorry for the hold up on merging this.  I need to make sure the Jackson version bump is okay for the rest of the code base.  At one point I needed to use an older Jackson library in render to be compatible with the older Spark environment we were using at Janelia.  There are now newer Spark environments available at Janelia so I should be able to get this all to work one way or another.  My goal is to get this merged ASAP so that it's available for Woods Hole.

igorpisarev:
I am also working on integrating my app with AWS S3 ‒ thanks @perlman, this PR was a great starting point for me!

Instead of intercepting calls to `Opener`, there is an option to register a custom `URLStreamHandlerFactory` at startup so S3 URLs work transparently. For details, please see https://github.com/saalfeldlab/stitching-spark/commit/11753c4e6d2a2b1ffb5737cc2907bc62bbc364c7

perlman:
Igor, that was my original approach, and I think I still have a compatible class in the commit.

I ran into issues having the property defined at run time with Jetty, relating to Java only allowing a single URLStreamHandlerFactory.

perlman:
Correction, I forgot one can use | on the property. I should revisit.

PR #46, opened 2017-10-06
TravisCI?

fcollman:
I wondered if you'd be interested in adding travisCI functionality to the repo.  This would let you see that pull requests have had tests run or not.  We could even extend it in the future to do more complicated integration tests.  This will put check marks next to every commit, and do it automatically for any PRs that come in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java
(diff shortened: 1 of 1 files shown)


trautmane:
Sure - I haven't used Travis CI (we have Jenkins setup for one of our larger projects here).  
I've been curious to explore/compare it.

PR #48, opened 2017-10-15
modified docker file to allow dynamic mongo configuration of render

fcollman:
I'm modified the render-ws docker to allow for dynamic configuration of the mongo database parameters via environment variables.  The entrypoint now uses sed to reconfigure the render-db.properties file with the MONGO_HOST and MONGO_PORT, which have defaults, and will do the same for authentication if appropriate environment variables are set. (MONGO_USERNAME,MONGO_PASSWORD).


fcollman:
i have a better more extensive PR

PR #73, opened 2018-01-08
spark exit: spark jobs hanging using local[*] master exit properly wi…

RussTorres:
…th explicit exit 0

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/ClientRunner.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/ClientRunner.java
@@ -37,6 +37,7 @@ public abstract class ClientRunner {
         try {
             runClient(args);
             LOG.info("run: exit, processing completed in {}", processTimer);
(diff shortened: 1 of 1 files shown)


RussTorres:
I haven't investigated the underlying problem here, but this blunt approach to cleaning up and exiting the spark jobs seems to work in most cases (at least using local[*]) which were not exiting and eating up wall time on our cluster.
There is likely a better solution, but it looks like the spark context is being stopped and rather than go deeper, I just tried this.

trautmane:
Hi @RussTorres, I'm guessing something Spark does with the local[*] setup can leave thread(s) running.  Your new System.exit(0) call forces all threads to exit.  I was thinking we shouldn't need to do this and was trying to come up with a case when this would be undesirable (e.g. if we really wanted a forked process to be able to run a little longer than the main thread of execution).  But I couldn't quickly come up with anything (and I was already explicitly exiting for exception / failure cases), so this change looks good to me.

PR #83, opened 2018-04-18
Sortable columns and last modified dates in stack listing

perlman:
...per a request from Forrest.


perlman:
There seems to be an issue with tablesort not updating correctly under safari.  I'm looking into it now.

perlman:
A few tweaks to properly clear the table contents on change, and to use long dates. These makes WebKit happy.

trautmane:
Thanks Eric!

Issue #76, opened 2018-01-31
Return png images as G or GA instead of RGBA?


perlman:
Some endpoints, e.g. /png-image, return data as RGBA (specifically, "PNG image data, 1024 x 1024, 8-bit/color RGBA, non-interlaced").

I am considering adding flags to force grayscale and/or disable transparency for PNG out-bit.  (The PNG format supports both G and GA at both 8- and 16-bit depths.)

Is this a reasonable approach?

perlman:
This is from playing around with support for 16-bit PNG output:
https://github.com/neurodata/render/blob/png16/render-ws/src/main/java/org/janelia/render/service/util/BufferedImageStreamingOutput.java

trautmane:
Hi Eric,  
Apologies for my delayed response.
This looks reasonable to me - kudos for digging into the https://github.com/leonbloy/pngj library details.
Did it work for you?
-- Eric T.

perlman:
The 16-bit grayscale PNG output is working well.

I have not touched 8-bit output yet. I don't want to break RGBA support for those that use it...

PR #100, opened 2019-06-24
16-bit PNG and raw image support

perlman:
I am reviving a long overdue thread for better png support (#76).

As of now, this branch adds a "png16-image" target, as well as "raw-image" (RGBA) and "raw16-image) (short gray), which I'm using for small experiment at the moment.

@trautmane: Instead of creating end-points for 16-bit, do you think the data type should be specified as an argument, e.g. &dtype=RGBA or &grayscale=true&depth=16 or something like that?

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/Utils.java
+++ b/render-app/src/main/java/org/janelia/alignment/Utils.java
@@ -62,6 +62,7 @@ public class Utils {
     public static final String PNG_FORMAT = "png";
     public static final String TIFF_FORMAT = "tiff";
     public static final String TIF_FORMAT = "tif";
+    public static final String RAW_FORMAT = "raw";
 
     private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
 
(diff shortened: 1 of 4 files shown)


trautmane:
Hi Eric,  Apologies for not looking at this sooner.  Your distinct end-points vs. query parameters question is one that I have struggled with for a while.  My original (possibly now outdated) rationale for creating distinct end-points for the image APIs was to ensure they were cache friendly.  Years ago some caches ignored query parameters.  Even today, I think most caches are sensitive to query parameter ordering (e.g.  ...grayscale=true&depth=16...  and ...depth=16&grayscale=true... reference the same image but get cached separately).  From a code standpoint, I think the use of query parameters is much more intuitive and maintainable.  Every time we introduce another image variant, I am enticed to move to the query parameter style.  Since you've already gone through the (mild) pain of adding distinct end-points here, let's keep what you've done.  I'm happy to consider/discuss alternatives if you have other thoughts on the subject.

trautmane:
Eric - I merged upstream changes and fixed one typo on a local copy of your branch.  Can you allow me to commit these changes to your repo ( see https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork ) so that I can easily merge this pull request with my fixes?

perlman:
"Allow edits from maintainers." is already checked.

These changes would be immediately useful for me to PR 16-bit support to neuroglancer, so, thanks! I will plan on adding arguments to reduce the number of endpoint permutations.  The png & raw endpoints should be able to handle all byte, short, RGB and RGBA with a small amount of work.

PR #148, opened 2023-09-22
Remove components related to abandoned 2018 hierarchical alignment ex…

trautmane:
…periments.  Rename HierarchicalDataService to CombinedRenderAndMatchService.  Add jakarta.annotation dependency.


render-ws/src/main/java/org/janelia/render/service/CombinedRenderAndMatchService.java
```diff
@@ -0,0 +1,256 @@
...
+
+            final List<TileBounds> allTileBounds = renderDao.getTileBoundsForZ(stackId, z, null);
+
+            final Set<String> distinctSectionIds = new HashSet<>(allTileBounds.size());
+            final Map<String, TileBounds> tileIdToBoundsMap = new HashMap<>(allTileBounds.size());
+
+            allTileBounds.forEach(tb -> {
+                final String sectionId = tb.getSectionId();
+                if (sectionId != null) {
+                    distinctSectionIds.add(tb.getSectionId());
```
minnerbe, line 103: This could be `distinctSectionIds.add(sectionId);`

render-ws/src/main/java/org/janelia/render/service/CombinedRenderAndMatchService.java
```diff
@@ -0,0 +1,256 @@
...
+                    tileIdToBoundsMap.put(tb.getTileId(), tb);
+                }
+            });
+
+            final Set<String> layerTileIds = tileIdToBoundsMap.keySet();
+            final TileIdsWithMatches tileIdsWithMatches = new TileIdsWithMatches();
+
+            final int numberOfSectionIds = distinctSectionIds.size();
+            if (numberOfSectionIds > 1) {
```
minnerbe, line 112: Sorting works with an empty set, and one or both of the loops get skipped if `distinctSectionIds` has one or zero elements, so the `if / else` branching shouldn't be necessary here.

minnerbe (approved):
Looks good to me! I only have two minor comments in `CombinedRenderAndMatchService` but addressing them is optional.

PR #149, opened 2023-09-25
Added flat interpolated model

minnerbe:
I added a model that interpolates an arbitrary number of `? extends Affine2d<?> & Model<?>`.
In comparison to the nested `InterpolatedModel2D` approach, the code is
1. much more readable and concise;
2. more general, since the new model can be constructed at runtime.

The new model is used in the alignment-solver. There are a bunch of tests that compare the new model to the old approach, and the [alignment of one layer of wafer 53](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B62782.75390625%2C44014%2C20.5%5D%2C%22crossSectionScale%22:1.4514872526305993%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafer_53/cut_000_to_009/c000_s095_v01_align_test_xy_qq%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22c000_s095_v01_align_test_xy_qq%22%7D%5D%2C%22selectedLayer%22:%7B%22layer%22:%22c000_s095_v01_align_test_xy_qq%22%7D%2C%22layout%22:%22xy%22%7D) looks fine. Still, it would be nice to compare the computed alignment with a previously computed one.
Is there any quick way to do this?


minnerbe:
I just validated the new model against the 'old' alignment (with `InterpolatedModel2D`). There are 5-10 (of 1729 tiles) that have a relative error of more than 5%, the rest is well below that (about 1% or even less).

This is the same for comparing two runs of the 'old' model on the same stack, since there is a considerable amount of randomness in the optimizer. Therefore, I conclude that this PR does not significantly change the alignment results.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/affine/AffineAlignBlockWorker.java
```diff
@@ -1035,6 +987,39 @@ else if ( preAlign == PreAlign.TRANSLATION )
...
+			for (final Tile<?> tile : tileConfig.getTiles()) {
+				final Tile<?> firstTile = solveItem.groupedTileToTiles().get(tile).get(0);
+				final String firstTileId = solveItem.tileToIdMap().get(firstTile);
+				final int z = (int) Math.round(solveItem.blockData().rtsc().getTileIdToSpecMap().get(firstTileId).getZ());
+				tileToZ.put(tile, z);
+			}
+			SolveTools.preAlignByLayerDistance(tileConfig, tileToZ);
+
+			errors = SolveTools.computeErrors(tileConfig.getTiles());
+			LOG.info("errors: " + errors);
```
trautmane, line 1016: see prior log suggestions

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/affine/AffineAlignBlockWorker.java
```diff
@@ -1035,6 +987,39 @@ else if ( preAlign == PreAlign.TRANSLATION )
...
+			// TODO: else they should be in the right position
+		} catch (final NotEnoughDataPointsException | IllDefinedDataPointsException e) {
+			LOG.info("block " + solveItem.blockData().getId() + ": prealign failed: ", e);
```
trautmane, line 1019: see prior log suggestions

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/BlockOptimizerParameters.java
```diff
@@ -66,23 +68,44 @@ public class BlockOptimizerParameters implements Serializable {
...
 	}
 
-	public InterpolatedAffineModel2D<InterpolatedAffineModel2D<InterpolatedAffineModel2D<AffineModel2D, RigidModel2D>, TranslationModel2D>, StabilizingAffineModel2D<RigidModel2D>>
-		getModel() {
+	public AlignmentModel getModel() {
+		final AlignmentModelBuilder builder = AlignmentModel.configure()
+				.addModel("affine", new AffineModel2D())
+				.addModel("rigid", new RigidModel2D())
+				.addModel("translation", new TranslationModel2D())
+				.addModel("regularization", new StabilizingAffineModel2D<>(new RigidModel2D()));
```
trautmane, line 83: make these string constants an enum (see earlier comment)

render-ws-java-client/src/test/java/org/janelia/render/client/AlignmentModelTest.java
```diff
@@ -100,6 +106,34 @@ public void interpolatingTheSameModelYieldsModel() {
...
+		matches.add(new PointMatch(new Point(new double[] { 1, 0 }), new Point(new double[] { 1, 0 })));
+		matches.add(new PointMatch(new Point(new double[] { 1, 1 }), new Point(new double[] { 1, 1.1 })));
+		matches.add(new PointMatch(new Point(new double[] { 0.5, 0.5 }), new Point(new double[] { 0.5, 0.5 })));
+
+		try {
+			interpolated.fit(matches);
+			model.fit(matches);
+		} catch (final NotEnoughDataPointsException | IllDefinedDataPointsException e) {
+			fail(e.getMessage());
+		}
```
trautmane, line 129: Add throws clause to the test method signature and remove the try/catch statements.  This reduces test code, but even better you'll get the exception message along with a stack-trace from JUnit.

trautmane (approved):
TLDR: Looks good - feel free to defer addressing refactoring comments until a later more convenient time.  

I decided I'd rather write down thoughts/suggestions now so we have them to discuss instead of limiting my comments to only those things I think must be addressed before merging.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/affine/AffineAlignBlockWorker.java
```diff
@@ -907,87 +907,41 @@ protected void solve(
...
+		LOG.info("errors: " + errors);
 
-				errors = SolveTools.computeErrors( tileConfig.getTiles() );
-				LOG.info("errors: " + errors);
-			}
-			// TODO: else they should be in the right position
-		} catch (final NotEnoughDataPointsException | IllDefinedDataPointsException e) {
-			LOG.info("block " + solveItem.blockData().getId() + ": prealign failed: ", e);
+		if (preAlignModelKey != null) {
+			preAlign(solveItem, tileConfig, preAlignModelKey);
```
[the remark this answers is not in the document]
minnerbe: Thanks! That really cleans up this part a lot.
I also really appreciate all the tips about logging, as I haven't done a lot of logging in the past.

PR #150, opened 2023-09-27
Optimize distributed solve match fetch

trautmane:
by adding resolvedTilesWithMatchesFrom API endpoint and integrating it into AffineAlignBlockWorker.  

This is just a first draft that produces equivalent results to the original distributed solve implementation.  Further refinement is expected.  The key optimization achieved was reduction in the number of web service requests for each solve and reduction in the amount of data returned for each request.  For a distributed solve of hess_wafer_53::c000_s095_v01, the total number of service GET requests were reduced from 5959 to 208.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
+++ b/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
@@ -57,9 +57,18 @@ public class ResolvedTileSpecsWithMatchPairs
      * Resolves all tile specs for client-side usage and normalizes match pairs
      * by removing pairs that are too far from each other in z and by sorting them.
      *
-     * @param  maxZDistance  maximum integral z distance for all retained pairs.
+     * @param  maxZDistance  maximum non-negative integral z distance for all retained pairs
+     *                       (or null to accept all pairs).
+     *
+     * @throws IllegalArgumentException
+     *   if maxZDistance < 0
      */
-    public void resolveTileSpecsAndNormalizeMatchPairs(final int maxZDistance) {
+    public void resolveTileSpecsAndNormalizeMatchPairs(final Integer maxZDistance)
+            throws IllegalArgumentException {
+
+        if ((maxZDistance != null) && (maxZDistance <= 0)) {
+            throw new IllegalArgumentException("maxZDistance must be >= 0 or null");
+        }
 
         resolvedTileSpecs.resolveTileSpecs();
 
@@ -69,9 +78,13 @@ public class ResolvedTileSpecsWithMatchPairs
             final TileSpec pTileSpec = resolvedTileSpecs.getTileSpec(pair.getpId());
             final TileSpec qTileSpec = resolvedTileSpecs.getTileSpec(pair.getqId());
             if ((pTileSpec != null) && (qTileSpec != null)) {
-                final int zDistance = (int) Math.abs(pTileSpec.getZ() - qTileSpec.getZ());
-                if (zDistance <= maxZDistance) {
+                if (maxZDistance == null) {
                     normalizedMatchPairs.add(pair);
+                } else {
+                    final int zDistance = (int) Math.abs(pTileSpec.getZ() - qTileSpec.getZ());
+                    if (zDistance <= maxZDistance) {
+                        normalizedMatchPairs.add(pair);
+                    }
                 }
             }
         }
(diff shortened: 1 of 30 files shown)


render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
```diff
@@ -30,37 +35,101 @@ public ResolvedTileSpecsWithMatchPairs(final ResolvedTileSpecCollection resolved
...
+        final Integer numberOfTilesToKeep = tileIdsToKeep == null ? null : tileIdsToKeep.size();
+
+        LOG.info("normalize: entry, process {} tiles and {} pairs with {} tileIdsToKeep and maxZDistance {}",
+                 resolvedTileSpecs.getTileCount(), matchPairs.size(), numberOfTilesToKeep, maxZDistance);
+
+        if ((maxZDistance != null) && (maxZDistance <= 0)) {
+            throw new IllegalArgumentException("maxZDistance must be >= 0 or null");
+        }
+
+        // remove unwanted tiles (which will later cause associated math pairs to be removed)
```
minnerbe, line 83: typo `math -> match`

render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
```diff
@@ -30,37 +35,101 @@ public ResolvedTileSpecsWithMatchPairs(final ResolvedTileSpecCollection resolved
...
+        if (tileIdsToKeep != null) {
+
+            // track and log removal info if debugging
+            final boolean isDebugEnabled = LOG.isDebugEnabled();
+            final Set<String> beforeTileIds = isDebugEnabled ? new HashSet<>(resolvedTileSpecs.getTileIds()) : null;
+
+            resolvedTileSpecs.removeDifferentTileSpecs(tileIdsToKeep);
```
minnerbe, line 90: The 'standard name' for this operation in the `Collection` interface would be `retainAll()`.

render-app/src/main/java/org/janelia/alignment/spec/TileSpec.java
```diff
@@ -118,6 +118,10 @@ public Double getZ() {
         return z;
     }
 
+    public Integer getIntegralZ() {
```
minnerbe, line 121: I would probably name this `getIntegerZ()` to remove ambiguities with the mathematical operation of integration.

render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
```diff
@@ -30,37 +35,101 @@ public ResolvedTileSpecsWithMatchPairs(final ResolvedTileSpecCollection resolved
...
+        final List<CanvasMatches> normalizedMatchPairs = new ArrayList<>(matchPairs.size());
+
+        for (final CanvasMatches pair : matchPairs) {
+            final TileSpec pTileSpec = resolvedTileSpecs.getTileSpec(pair.getpId());
+            final TileSpec qTileSpec = resolvedTileSpecs.getTileSpec(pair.getqId());
+            if ((pTileSpec != null) && (qTileSpec != null)) {
+                if (maxZDistance == null) {
                     normalizedMatchPairs.add(pair);
+                } else {
+                    final int zDistance = (int) Math.abs(pTileSpec.getZ() - qTileSpec.getZ());
```
minnerbe, line 117: Could this use the new `getIntegralZ()` method?

minnerbe:
Other than the minor comments above, seems good to go. The reduction of the queries by an order of magnitude is great!

Thanks for cleaning up the `AffineAlignBlockWorker` a bit. I'm currently also cleaning this up while trying to split input and output, so I might have to rebase my branch, but this shouldn't be a big deal.

render-app/src/main/java/org/janelia/alignment/spec/TileSpec.java
trautmane: good point - done

render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
trautmane: Yeah - I did not realize that years ago when I wrote the original method - doh.  I've updated the name.

render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
trautmane: done

render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
trautmane: yes - actually decided to take it one step further and add a zDistanceFrom method to TileSpec instead

PR #152, opened 2023-09-28
Try to separate output from inputs

minnerbe:
I encapsulated most of what is computed during a run in the `ResultContainer` object. Most of its fields are strictly encapsulated (except for `idToModel`, which is sometimes used for iterating). A result container is built by getting a `ResolvedTileSpecCollection` and then records all needed quantities through clear method calls.


trautmane:
merging locally and removing branch

PR #154, opened 2023-10-26
Make intensities uniform again

minnerbe:
This is an attempt at providing a parameter to adjust the overall uniformity during intensity correction.

**Before**: Coefficient tiles within an image tile were identity connected, parts that overlap with another image tile are matched. This ensures global smoothness of the result (i.e., no hard cuts between two image tiles).

**Now**: In addition, all coefficient tiles are connected to a single tile by a point match `"average intensity of coefficient tile" <-> 0.5`. This tile is fixed during optimization. This should skew the average intensity of every tile towards the "target intensity" `0.5`. A parameter `--equilibrationWeight` is provided to tune the weight of this match; setting this parameter to `0.0` disables this mechanism and does intensity correction as before.


trautmane (commented):
Looks fine except for compilation failure after merging into refactor/separate-output.  I'm guessing this should be merged into ibeam_msem first and then ibeam_msem can be merged into refactor/separate-output.  I'll stop by to discuss ...

trautmane:
Needed to merge this into newsolver branch locally - changing the branch target in the GutHub UIO was a bad idea.

PR #155, opened 2023-10-27
Changing block layout for alignment

minnerbe:
I changed the way the XY block layout is chosen. Starting from the regular grid:

```
-----------
|    |    |
|----|----|
|    |    |
-----------
```
* The old code placed a block of the same size in the middle of this grid. All tiles that touched a block were pulled from the DB and aligned. Only tiles that had their midpoint completely within a block were kept for global alignment. This lead to 1) no overlap at the boundaries of the grid; 2) a relatively small layer of tiles that were not kept after alignment (which is bad because the boundary tiles could be aligned badly).
* The new code simply grows the blocks by a factor of 2 in every dimension, which guarantees sufficient overlap. After block-alignment and before global interpolation, the outermost 25% in every dimension is shaved off.


trautmane (approved):
Nice!

PR #159, opened 2023-11-21
Background correction for cellmap stack

minnerbe:
I started implementing background correction as discussed:
1. average all affected tiles in the stack
2. heavily smooth the resulting image
3. subtract the resulting background image from all affected tiles (as `Filter`)

However, I see some problems with this approach (technical and conceptional):
* Currently, the `Filter.process` method only gets an `ImageProcessor`, so there is no way of matching the incoming image to the background in terms of stage coordinates.
* As is apparent from [this picture](https://github.com/JaneliaSciComp/recon_fibsem/issues/47#issuecomment-1731889642), the light region in the stack varies in spatial extent and intensity throughout the stack. Simply subtracting one background image from everything probably doesn't completely get rid of the problem.

**Alternative**: we could use the already implemented `RollingBallSubtraction` filter that uses an ImageJ plugin to do local background correction for a single image. The results of applying this filter don't look too bad (original vs. rolling paraboloid vs. rolling ball):
original-vs-paraboloid-vs-ball

The downside would be that this is pretty computation heavy, so pre-applying the filter would be necessary, essentially duplicating the whole stack.
What are your thoughts @StephanPreibisch ?

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/filter/AverageBackgroundSubtraction.java
+++ b/render-app/src/main/java/org/janelia/alignment/filter/AverageBackgroundSubtraction.java
@@ -1,0 +1,54 @@
+package org.janelia.alignment.filter;
+
+import ij.process.ImageProcessor;
+import mpicbg.trakem2.transform.TransformMeshMappingWithMasks;
+import org.janelia.alignment.loader.ImageLoader;
+import org.janelia.alignment.util.ImageProcessorCache;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class AverageBackgroundSubtraction implements Filter {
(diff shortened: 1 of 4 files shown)


trautmane:
> Currently, the Filter.process method only gets an ImageProcessor, so there is no way of matching the incoming image to the background in terms of stage coordinates.

Yeah - I guess to make this work, we'd have to inversely transform the full background image and chop it up into separate tiles.  So, you'd end up with 4 background tiles for the thymus data set (2 rows x 2 columns).  Not sure if this is doable.

trautmane:
I guess I'll go ahead and close this now ...

PR #162, opened 2023-11-29
Alternative background correction

minnerbe:
This is an alternative to #159 as discussed in person. Rather than pre-computing one background image for the whole stack, I added a `Filter` implementation that, for every image, does the following:
* convert picture to 32-bit grayscale
* downsample (by factor of twenty in order to make the subsequent median filtering sufficiently fast)
* median filtering with mirrored boundary (with a 1000px radius on the full-res image)
* subtract mean from the filtered background
* subtract the (upsampled) filtered background from the original image
* convert to 8-bit grayscale again

### Results

On my laptop, correcting one image takes about 3-5sec, so this shouldn't be too costly to do for the whole stack. The corrected images look much nicer than even the rolling ball method. In particular, note that there is no shadow in the resin area at the bottom of the images (top is after correction):

image_0-1-1_1250_after
image_0-1-1_1250_before

Furthermore, this method seems to not clip any values, but to smooth the histogram  (top is after correction):
histogram_0-1-1_1250_after
histogram_0-1-1_1250_before

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/BackgroundCorrectionClient.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/BackgroundCorrectionClient.java
@@ -8,7 +8,6 @@ import ij.measure.Measurements;
 import ij.plugin.ImageCalculator;
 import ij.plugin.Scaler;
 import ij.plugin.filter.RankFilters;
-import ij.process.ImageConverter;
 import ij.process.ImageProcessor;
 import ij.process.ImageStatistics;
 import net.imglib2.RandomAccessibleInterval;
@@ -109,11 +108,11 @@ public class BackgroundCorrectionClient {
 			final ImageProcessor ip = loadImage(tileSpec, imageProcessorCache);
 
 			final long start = System.currentTimeMillis();
-			final ImageProcessor processedImage = subtractBackground(ip);
+			subtractBackground(ip);
 			final long end = System.currentTimeMillis();
 			LOG.info("Corrected background for tile {} in {} ms", tileSpec.getTileId(), end - start);
 
-			saveImage(processedImage, tileSpec);
+			saveImage(ip, tileSpec);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


StephanPreibisch:
looks great!

render-ws-java-client/src/main/java/org/janelia/render/client/BackgroundCorrectionClient.java
```diff
@@ -0,0 +1,156 @@
...
+import java.nio.file.Path;
+
+/**
+ * Create and store a background image that can be used to subtract a
+ * spatially varying background from a set of images.
+ * Originally developed for the CellMap jrc_mus_thymus_1 dataset.
+ *
+ * @author Michael Innerberger
+ */
+public class BackgroundCorrectionClient {
```
trautmane, line 31: With a deployed named filter configuration ( see [some examples here](http://renderer-dev.int.janelia.org:8080/render-ws/v1/namedFilterSpecLists) ) and `--filterListName ... --excludeAllTransforms` parameters, I think the [RenderTilesClient](https://github.com/saalfeldlab/render/blob/newsolver/render-ws-java-client/src/main/java/org/janelia/render/client/RenderTilesClient.java#L51-L59) would cover everything done by this client (along with some other features this client doesn't have).  I plan to use the RenderTilesClient for the thymus data set.  

The one nice thing this client has over the RenderTilesClient is that you don't need to deploy a named filter because you specify the filter parameters on the command line.  I'm waffling on whether we should:

a) keep this client and add a TODO to look at a good way to integrate this convenience into the RenderTileClient or 
b) simply remove this client as an almost-duplicate to reduce future maintenance.

Let me know what you think.

StephanPreibisch:
nice idea to downsample ... might blur the edge mask a little, but the effect seems neglectable ...

trautmane (approved):
Nice!  See my inlined comments ...

minnerbe:
The downsampling was necessary since the median filtering needs a pretty large radius (almost 20% of the original image size) to yield good results. When trying to do this in Fiji at full resolution, my guesstimate would be 10-30min per image.

I'm not completely sure what the edge mask is, but the background correction does seem to be slightly worse at some parts of the image boundary.

render-ws-java-client/src/main/java/org/janelia/render/client/BackgroundCorrectionClient.java
minnerbe: I'd go for b. This client was just a means for me to test the whole process. If there is an existing client that can be used instead, we can definitely delete the new client.

PR #163, opened 2023-12-07
Fix duplicate tile pairs bug

trautmane:
when deriving an OrderedCanvasIdPair withRelativePositions from two tiles with the same bounds.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
+++ b/render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
@@ -8,8 +8,8 @@ import org.janelia.alignment.json.JsonUtils;
 
 import jakarta.annotation.Nonnull;
 
-import static org.janelia.alignment.match.MontageRelativePosition.LEFT;
-import static org.janelia.alignment.match.MontageRelativePosition.TOP;
+import static org.janelia.alignment.Utils.NULLS_FIRST_STRING_COMPARATOR;
+import static org.janelia.alignment.match.MontageRelativePosition.*;
 
 /**
  * Key identifiers for a canvas.
@@ -170,8 +170,7 @@ public class CanvasId
     public static final double[] ZERO_OFFSETS = { 0.0, 0.0 };
 
     private static final Comparator<CanvasId> CANVAS_ID_COMPARATOR = Comparator
-            .comparing(CanvasId::getGroupId, Comparator.nullsFirst(Comparator.naturalOrder()))
-            .thenComparing(CanvasId::getId, Comparator.nullsFirst(Comparator.naturalOrder()))
-            .thenComparing(CanvasId::getRelativePosition, Comparator.nullsFirst(Comparator.naturalOrder()));
-
+            .comparing(CanvasId::getGroupId, NULLS_FIRST_STRING_COMPARATOR)
+            .thenComparing(CanvasId::getId, NULLS_FIRST_STRING_COMPARATOR)
+            .thenComparing(CanvasId::getRelativePosition, NULLS_FIRST_POSITION_COMPARATOR);
 }
--- a/render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
+++ b/render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
@@ -1,9 +1,12 @@
 package org.janelia.alignment.match;
 
 import java.io.Serializable;
+import java.util.Comparator;
 import java.util.Objects;
 
 import org.janelia.alignment.json.JsonUtils;
+
+import jakarta.annotation.Nonnull;
 
 import static org.janelia.alignment.match.MontageRelativePosition.LEFT;
 import static org.janelia.alignment.match.MontageRelativePosition.TOP;
@@ -116,6 +119,10 @@ public class CanvasId
         }
     }
 
+    public CanvasId withRelativePosition(final MontageRelativePosition relativePosition) {
+        return new CanvasId(groupId, id, relativePosition);
+    }
+
     public CanvasId withoutRelativePosition() {
         return new CanvasId(groupId, id);
     }
@@ -136,23 +143,12 @@ public class CanvasId
 
     @Override
     public int hashCode() {
-        return java.util.Objects.hash(id, groupId, relativePosition);
+        return Objects.hash(id, groupId, relativePosition);
     }
 
     @Override
-    public int compareTo(final CanvasId that) {
-        int result = this.groupId.compareTo(that.groupId);
-        if (result == 0) {
-            result = this.id.compareTo(that.id);
-            if (result == 0) {
-                if (this.relativePosition == null) {
-                    result = that.relativePosition == null ? 0 : -1;
-                } else if (that.relativePosition != null) {
-                    result = this.relativePosition.compareTo(that.relativePosition);
-                }
-            }
-        }
-        return result;
+    public int compareTo(@Nonnull final CanvasId that) {
+        return CANVAS_ID_COMPARATOR.compare(this, that);
     }
 
     @Override
@@ -172,5 +168,10 @@ public class CanvasId
             new JsonUtils.Helper<>(CanvasId.class);
 
     public static final double[] ZERO_OFFSETS = { 0.0, 0.0 };
+
+    private static final Comparator<CanvasId> CANVAS_ID_COMPARATOR = Comparator
+            .comparing(CanvasId::getGroupId, Comparator.nullsFirst(Comparator.naturalOrder()))
+            .thenComparing(CanvasId::getId, Comparator.nullsFirst(Comparator.naturalOrder()))
+            .thenComparing(CanvasId::getRelativePosition, Comparator.nullsFirst(Comparator.naturalOrder()));
 
 }
--- a/render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
+++ b/render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
@@ -3,6 +3,8 @@ package org.janelia.alignment.match;
 import java.io.Serializable;
 import java.util.Objects;
 
+import org.janelia.alignment.spec.TileBounds;
+
 /**
  * A pair of canvas identifiers with {@linkplain Comparable natural ordering}.
  *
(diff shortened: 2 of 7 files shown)


render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
```diff
@@ -55,6 +57,47 @@ public OrderedCanvasIdPair(final CanvasId oneId,
...
+     * @return an ordered pair where each CanvasId includes relative position information
+     *         based upon the specified tile bounds.
+     *         Assumes that the bounds are for the same layer so the pair's deltaZ is set to null.
+     *
+     * @throws IllegalArgumentException
+     *   if both tile identifiers are the same.
+     */
+    public static OrderedCanvasIdPair withRelativePositions(final TileBounds oneTileBounds,
+                                                            final TileBounds anotherTileBounds)
+            throws IllegalArgumentException {
```
minnerbe, line 73: It seems to me that there is some redundancy or even misplaced responsibility in the classes used here:
* An `OrderedCanvasIdPair` can be constructed by using `TileBounds` as well as `CanvasId`, which was surprising to me because I didn't expect `TileBounds` to carry an id. If that is to be desired, I would suggest letting `TileBounds` store a `CanvasId` with an appropriate getter.
* `CanvasId` carries a relative geometric information in xy-direction, whereas `OrderedCanvasIdPair` contains a relative position in z-direction. Both geometric informations seem to be optional. I would suggest encapsulating relative geometric information into one class and only store a reference to an instance in `OrderedCanvasIdPair`.

Keep in mind: those are just suggestions. There might be better ways to improve the issues mentioned, or there might be arguments against changing the classes at all. Furthermore, this doesn't need to be addressed within this PR.

render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
```diff
@@ -55,6 +57,47 @@ public OrderedCanvasIdPair(final CanvasId oneId,
...
+
+        final CanvasId oneCanvasId = new CanvasId(oneTileBounds.getSectionId(), oneTileBounds.getTileId());
+        final CanvasId anotherCanvasId = new CanvasId(anotherTileBounds.getSectionId(), anotherTileBounds.getTileId());
+
+        final MontageRelativePosition[] positions = MontageRelativePosition.getRelativePositions(oneTileBounds,
```
minnerbe, line 78: It seems to me that the signature of `MontageRelativePosition.getRelativePosition` (which I would rename to `MontageRelativePosition.of` and only let it return one position instead of an array) is pretty much tailored to this use here and that some of the responsibility (e.g., checking if two `TileBounds` are the same -- maybe introduce `MontageRelativePosition.SAME`?) depend on the implementation within `MontageRelativePosition` and therefore should go there.

render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
```diff
@@ -55,6 +57,47 @@ public OrderedCanvasIdPair(final CanvasId oneId,
...
+        // if the tile identifiers are not in natural order ...
+        if (comparisonResult > 0) {
+            final MontageRelativePosition[] flippedPositions =
+                    MontageRelativePosition.getRelativePositions(anotherTileBounds,
+                                                                 oneTileBounds);
+            // and if both tiles are in the same position relative to each other ...
+            if (positions[0].equals(flippedPositions[0])) {
+                // then flip the positions to maintain consistency with natural order
+                positions[0] = flippedPositions[1]; // NOTE: this is flipping because both are same
+                positions[1] = flippedPositions[0];
```
minnerbe, line 90: I see that swapping `positions` without an auxiliary variable is not possible here and assigning from `flippedPositions` is convenient. However, from the naming it seems at first glance that this does **not** flip the array items.

render-app/src/main/java/org/janelia/alignment/spec/TileBoundsRTree.java
```diff
@@ -304,21 +302,16 @@ public static Set<OrderedCanvasIdPair> getDistinctPairs(final TileBounds fromTil
...
 
+                        final OrderedCanvasIdPair pair;
                         if (includeRelativePosition) {
-                            relativePositions = MontageRelativePosition.getRelativePositions(fromTile, toTile);
-                            pairs.add(
-                                    new OrderedCanvasIdPair(
-                                            new CanvasId(fromTile.getSectionId(), pTileId, relativePositions[0]),
-                                            new CanvasId(toTile.getSectionId(), qTileId, relativePositions[1]),
-                                            null));
+                            pair = OrderedCanvasIdPair.withRelativePositions(fromTile, toTile);
```
minnerbe, line 307: Also here: if I set the `includeRelativePosition` flag, I would expect that the relative z-distance is included.

render-app/src/test/java/org/janelia/alignment/spec/TileBoundsRTreeTest.java
```diff
@@ -220,4 +240,45 @@ private List<TileBounds> buildListForZ(final double z) {
 
         return list;
     }
+
+    @SuppressWarnings("SameParameterValue")
+    private List<OrderedCanvasIdPair> buildExpectedPairs(final double pZ,
+                                                         final int pTileIndex,
+                                                         final double qZ,
+                                                         final int... qTileIndexes) {
```
minnerbe, line 248: This is super specific, but: when referring to the mathematical meaning, the plural of index is indices. In contrast, the plural form indexes is used when referring to databases or documents.

minnerbe (commented):
I think the overall goal of catching the case of two tiles with the same bounds is achieved by this PR. I suggest to make this explicit and test this by trying to create an `OrderedCanvasIdPair` with the same tile. Unfortunately, `Assert.assertThrows` was only introduced in JUnit5.
Furthermore, I have some additional concerns about responsibilities and boundaries of some of the included classes. Those don't need to be addressed in this PR, but we should at least make a TODO and/or discuss these concerns.

render-app/src/test/java/org/janelia/alignment/match/OrderedCanvasIdPairTest.java
```diff
@@ -22,8 +23,49 @@ public void testNormalization() {
...
+                                                  boundsA.getMinX() + 5.0, boundsA.getMinY(),
+                                                  boundsA.getMaxX() + 5.0, boundsA.getMaxY());
+
+        final OrderedCanvasIdPair leftRightPair = OrderedCanvasIdPair.withRelativePositions(boundsA,
+                                                                                            boundsB);
+        validatePair("leftRightPair", leftRightPair, tileAId, MontageRelativePosition.LEFT);
+
+        final String tileOverlapId = "tileOverlap";
+        final TileBounds boundsOverlap = new TileBounds(tileOverlapId, boundsA.getSectionId(), boundsA.getZ(),
+                                                        boundsA.getMinX(), boundsA.getMinY(),
```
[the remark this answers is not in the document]
trautmane: good idea

render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
```diff
@@ -173,4 +169,9 @@ public String toJson() {
 
     public static final double[] ZERO_OFFSETS = { 0.0, 0.0 };
 
+    private static final Comparator<CanvasId> CANVAS_ID_COMPARATOR = Comparator
+            .comparing(CanvasId::getGroupId, Comparator.nullsFirst(Comparator.naturalOrder()))
+            .thenComparing(CanvasId::getId, Comparator.nullsFirst(Comparator.naturalOrder()))
+            .thenComparing(CanvasId::getRelativePosition, Comparator.nullsFirst(Comparator.naturalOrder()));
+
```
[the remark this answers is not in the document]
trautmane: I committed a static import variant.  I'm not convinced it is a big improvement, but it's fine.

render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
trautmane: Yeah - I see what you mean.  In retrospect, I should have looked into merging the concepts when they evolved.  Here is how/why (I think) I created the spaghetti: `TileBounds` were created to couple tile context (tileId and sectionId) with bounds info to support lighter-weight APIs where transformation details were not needed.  The `MontageRelativePosition` was added to `CanvasId` when we started clipping tiles to optimize point match derivation for same z-layer tiles.  The `absoluteDeltaZ` was added to `OrderedCanvasIdPair` when I introduced multi-stage matching which needed to toss too-distant pairs in later stages.

I think you have the right idea to encapsulate all relative positional information for a pair into a single class.  I'm adding a TODO for this since I don't have time to clean this up right now.

render-app/src/main/java/org/janelia/alignment/match/OrderedCanvasIdPair.java
trautmane: tried to improve naming

render-app/src/main/java/org/janelia/alignment/match/CanvasId.java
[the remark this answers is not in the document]
minnerbe: I do like how this reads now. It's not about big improvements, but I have the feeling that I'd get the idea much quicker now if looking at that snippet for the first time.

minnerbe:
Just had a look at all the changes you pushed. I think they look great!
I also agree that we should defer refactoring the position classes to a point where we have enough time to do it properly.

From my side, this PR is ready to be merged.

render-app/src/main/java/org/janelia/alignment/spec/TileBoundsRTree.java
trautmane: makes sense - renamed this to make it clearedr

render-app/src/test/java/org/janelia/alignment/spec/TileBoundsRTreeTest.java
trautmane: Ahh - I concede defeat on this one 🥲

PR #165, opened 2023-12-19
alternating intensity correction fixes

trautmane:
Let's accumulate and review fixes here ...


minnerbe:
Looks good to me. I confirmed that the fix worked on a small stack. I think we can merge that into the feature branch so that we can further develop the alternating solve for intensity correction.

minnerbe (approved):

PR #160, opened 2023-11-22
Alternating block solve

minnerbe:
### Changes

I added a client that does a couple of rounds of the distributed block solver with alternating layouts to ensure that every tile is in the "interior" of a block at least once. More specifically, the algorithm looks like this:
1. solve blocks with overlap
2. instead of interpolating transforms on overlapping regions, choose transform randomly from all candidates
3. save blocks to database with coherent naming scheme (`_runN`)
4. change parameters:
   - change stack to pull data from to target stack of previous iteration
   - shift block layout by half a block (before expanding, i.e., by 1/4 of an expanded block)
5. go to 1.

### Test

I tested this on a small toy stack (1 mFOV):
- For comparison: [here](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B43877%2C9212%2C424%5D%2C%22crossSectionScale%22:16%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafer_53/cut_000_to_009/c009_s310_v01_mfov_08_exact%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22c009_s310_v01_mfov_08_exact%22%7D%5D%2C%22selectedLayer%22:%7B%22layer%22:%22c009_s310_v01_mfov_08_exact%22%7D%2C%22layout%22:%22xy%22%7D) is a stack that was solved globally
[...]

PR #166, opened 2023-12-21
Alternating solver for intensity correction

minnerbe:
In the spirit of the alternating solver for alignment (see #160), I added a similar solver for intensity correction. It works in exactly the same way as the analogue for alignment, but two changes to the existing code were necessary:
- The rendering method that is used to match intensities between tiles didn't take into account existing intensity filters. It does now. I put this into `Render.render` rather than `VisualizeTools.getUntransformedProcessorWithMasks` to preserve my notion of "untransformed" (meaning also no filters).
- A `LinearIntensityMap8BitFilter` can now absorb a second filter to emulate chaining of linear functions.

### Results
I tested the implementation on one layer of wafer_53 with a 3x3 XY-block layout; see pictures below.
While I can still see some tile boundaries after one run, the results after two runs look qualitatively very close to the baseline (intensity correction without blocks).

- Original layer: ng-no-ic
- Reference IC (no blocks): ng-baseline
- One run of blocked IC: ng-1run
- Two consecutive runs of blocked IC: ng-2runs

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/filter/FilterSpec.java
+++ b/render-app/src/main/java/org/janelia/alignment/filter/FilterSpec.java
@@ -30,7 +30,7 @@ public class FilterSpec {
     private final String className;
     private final Map<String, String> parameters;
 
-    private transient Class clazz;
+    private transient Class<?> clazz;
 
     // no-arg constructor needed for JSON deserialization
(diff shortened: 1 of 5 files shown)


trautmane (approved):
looks great!

PR #167, opened 2024-01-07
newsolver alignment pipeline v2

trautmane:
I started refactoring the spark alignment pipeline components to simplify future development.  The first draft of the pipeline utilized a JSON configuration file in conjunction with a hard-coded ordering of components.  This draft introduces an AlignmentPipelineStep interface that clarifies what a pipeline step must support and allows step ordering to be specified in the JSON configuration file.  

If/when we are happy with this draft, the plan is to add intensity correction and thickness correction steps to the pipeline.

For review, I recommend starting by looking first at the [AlignmentPipelineStep interface](https://github.com/saalfeldlab/render/blob/ad30a5ed8d9488292da9d38c0bada88755b7f812/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStep.java) and how use of it simplifies the AlignmentPipelineClient.  There is also an [test/example JSON file](https://github.com/saalfeldlab/render/blob/ad30a5ed8d9488292da9d38c0bada88755b7f812/render-ws-spark-client/src/test/resources/pipeline/msem_alignment_pipeline.json) that might be useful.

### DIFF
--- a/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/ClusterCountClient.java
+++ b/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/ClusterCountClient.java
@@ -16,7 +16,7 @@ import org.janelia.alignment.spec.stack.StackId;
 import org.janelia.alignment.spec.stack.StackWithZValues;
 import org.janelia.render.client.ClientRunner;
 import org.janelia.render.client.RenderDataClient;
-import org.janelia.render.client.parameter.AlignmentPipelineParameters;
+import org.janelia.render.client.spark.pipeline.AlignmentPipelineParameters;
 import org.janelia.render.client.parameter.CommandLineParameters;
 import org.janelia.render.client.parameter.MultiProjectParameters;
 import org.janelia.render.client.parameter.TileClusterParameters;
--- a/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/ClusterCountClient.java
+++ b/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/ClusterCountClient.java
@@ -4,6 +4,7 @@ import com.beust.jcommander.ParametersDelegate;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.spark.SparkConf;
@@ -16,11 +17,12 @@ import org.janelia.alignment.spec.stack.StackId;
 import org.janelia.alignment.spec.stack.StackWithZValues;
 import org.janelia.render.client.ClientRunner;
 import org.janelia.render.client.RenderDataClient;
-import org.janelia.render.client.spark.pipeline.AlignmentPipelineParameters;
 import org.janelia.render.client.parameter.CommandLineParameters;
 import org.janelia.render.client.parameter.MultiProjectParameters;
 import org.janelia.render.client.parameter.TileClusterParameters;
 import org.janelia.render.client.spark.LogUtilities;
+import org.janelia.render.client.spark.pipeline.AlignmentPipelineParameters;
+import org.janelia.render.client.spark.pipeline.AlignmentPipelineStep;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,7 +32,7 @@ import org.slf4j.LoggerFactory;
  * @author Eric Trautman
  */
 public class ClusterCountClient
-        implements Serializable {
+        implements Serializable, AlignmentPipelineStep {
 
     public static class Parameters extends CommandLineParameters {
         @ParametersDelegate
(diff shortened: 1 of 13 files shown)


render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStep.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Validates the specified pipeline parameters are sufficient for this step.
+     *
+     * @param  pipelineParameters  parameters to validate.
+     *
+     * @throws IllegalArgumentException
+     *   if any of the parameters are invalid or missing.
+     */
+    void validatePipelineParameters(final AlignmentPipelineParameters pipelineParameters)
+            throws IllegalArgumentException;
```
minnerbe, line 23: There are a few points that are not clear to me when I try to imagine how this method is used:
- Can it mutate the given `AlignmentPipelineParameters` or not? Either way, this should be made explicit, e.g. by simply stating it in the documentation, or with a `@Contract(pure = true)` annotation.
- The method does not return anything, so only the the "negative" outcome leads to observable behavior. Maybe returning a `bool` would be better? But then, `false` is redundant...

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStepId.java
```diff
@@ -0,0 +1,46 @@
...
+    GENERATE_MIPMAPS,
+    DERIVE_TILE_MATCHES,
+    PATCH_MFOV_MONTAGE_MATCHES,
+    FIND_UNCONNECTED_MFOVS,
+    FIND_UNCONNECTED_TILES_AND_EDGES,
+    FILTER_MATCHES,
+    ALIGN_TILES;
+
+    public AlignmentPipelineStep toStepClient()
+            throws UnsupportedOperationException {
```
minnerbe, line 27: I usually like to give the enum a private field of type `Supplier<T>` instead of a huge switch statement. This requires a bit more effort up front, but makes the `toStepClient` method generic, such that adding a new enum doesn't have to be done in two places. E.g., define a new enum as `GENERATE_MIPMAPS(MipmapClient::new)` and have the `toStepClient` method only do
```Java
this.supplier.get()
```
Then, you can also get rid of the `UnsupportedOperationException`, since this case cannot happen anymore.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParameters.java
```diff
@@ -63,62 +73,48 @@ public MultiProjectParameters getMultiProject() {
         return multiProject;
     }
 
-    public boolean hasMipmapParameters() {
-        return mipmap != null;
-    }
-
```
minnerbe, line 69: Why did you delete all those `has...` methods? Just asking to understand.
Edit: saw that these were only used in `AlignmentPipelineClient` to check if the pipeline step is to be executed. I like the current design much more!

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParameters.java
```diff
@@ -63,62 +73,48 @@ public MultiProjectParameters getMultiProject() {
...
     public AffineBlockSolverSetup getAffineBlockSolverSetup() {
         return affineBlockSolverSetup;
     }
 
+    public List<AlignmentPipelineStep> buildStepClients()
+            throws IllegalArgumentException {
+
+        if ((pipelineSteps == null) || pipelineSteps.isEmpty()) {
+            throw new IllegalArgumentException("no pipeline steps defined");
+        }
```
minnerbe, line 109: Just to make things clear: we go for "fail fast" rather than a smooth execution (in which case we could return an empty list), right? In this case, the check for missing parameters should probably be done right after JSON deserialization.
The exception should also be noted in the doc-string for this method.

render-ws-spark-client/src/test/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParametersTest.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class AlignmentPipelineParametersTest {
+
+    @Test
+    public void testJson()
+            throws IOException {
+        final AlignmentPipelineParameters pipelineParameters =
+                AlignmentPipelineParameters.fromJsonFile("src/test/resources/pipeline/msem_alignment_pipeline.json");
+
+        Assert.assertNotNull("deserialized parameters are null", pipelineParameters);
```
minnerbe, line 23: We should consider switching to JUnit5 in the long run: it always confuses me if I read something like "assertNotNull(".. are null")". JUnit5 places the `msg` parameter at the end of the call.

render-ws-spark-client/src/test/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParametersTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+        Assert.assertEquals("incorrect affineBlockSolverSetup.targetStack.stackSuffix value parsed",
+                            "_align_pipe_alt_aa", affineBlockSolverSetup.targetStack.stackSuffix);
+
+        Assert.assertTrue("incorrect affineBlockSolverSetup.targetStack.completeTargetStack value parsed",
+                          affineBlockSolverSetup.targetStack.completeStack);
+
+        final List<AlignmentPipelineStep> stepClients = pipelineParameters.buildStepClients();
+        Assert.assertNotNull("stepClients is null", stepClients);
+        Assert.assertEquals("incorrect number of stepClients", 7, stepClients.size());
```
minnerbe, line 36: This test method does a lot of things:
- checking if JSON deserialization produces a non-null object;
- verify that the correct target stack settings are loaded;
- testing the conversion to `AlignmentPipelineStep`.
Should we maybe split this into several tests with more descriptive names? This way the test could also serve as kind of documentation.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/ClusterCountClient.java
```diff
@@ -46,63 +48,80 @@ public org.janelia.render.client.ClusterCountClient buildJavaClient() {
...
+    public void validatePipelineParameters(final AlignmentPipelineParameters pipelineParameters)
+            throws IllegalArgumentException {
+        AlignmentPipelineParameters.validateRequiredElementExists("tileCluster",
+                                                                  pipelineParameters.getTileCluster());
     }
 
-    public void run() throws IOException {
+    /** Run the client as part of an alignment pipeline. */
+    public void runPipelineStep(final JavaSparkContext sparkContext,
+                                final AlignmentPipelineParameters pipelineParameters)
```
minnerbe, line 90: I really like how the changed handling of the parameter object now makes the steps seem like proper pipeline steps (the parameters now act as shared context instead of internal state of the steps).

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/CopyMatchClient.java
```diff
@@ -47,56 +49,63 @@ public Parameters(final MultiProjectParameters multiProject,
...
 
-    public void copyMatches(final JavaSparkContext sparkContext)
-            throws IOException {
+    /** Validates the specified pipeline parameters are sufficient. */
+    @Override
+    public void validatePipelineParameters(final AlignmentPipelineParameters pipelineParameters)
+            throws IllegalArgumentException {
+        AlignmentPipelineParameters.validateRequiredElementExists("matchCopy",
+                                                                  pipelineParameters.getMatchCopy());
+    }
```
minnerbe, line 88: The validation seems to only verify that the parameters are present, not that they are valid. Is the name maybe too general for the purpose it serves?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParameters.java
```diff
@@ -50,6 +59,7 @@ public AlignmentPipelineParameters(final MultiProjectParameters multiProject,
                                        final MatchCopyParameters matchCopy,
                                        final AffineBlockSolverSetup affineBlockSolverSetup) {
```
minnerbe, line 60: This hardcodes all existing steps, isn't it? Is there a way around that? I guess every approach that deals with those in a more generic (and dynamic) way probably loses type information, which would be a strong argument against doing that.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/MFOVMontageMatchPatchClient.java
```diff
@@ -43,84 +44,86 @@ public static class Parameters extends CommandLineParameters {
...
             @Override
             public void runClient(final String[] args) throws Exception {
-
                 final Parameters parameters = new Parameters();
                 parameters.parse(args);
-
-                final MFOVMontageMatchPatchClient client = new MFOVMontageMatchPatchClient(parameters);
-                client.run();
+                final MFOVMontageMatchPatchClient client = new MFOVMontageMatchPatchClient();
+                client.createContextAndRun(parameters);
```
minnerbe, line 57: Does every client create its own context? If so, do we want that (I'm not familiar enough with Spark contexts to answer this question myself)?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStep.java
trautmane: I tried to add some more-helpful javadocs.  Bottom line is no mutation and throw exception for problem.  This is a validation pattern/approach that makes sense to me and is repeated in many other contexts.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStepId.java
trautmane: Great idea - done.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParameters.java
trautmane: Yes - fail fast.  This check is effectively done right after deserialization.  Added javadoc - good point.

render-ws-spark-client/src/test/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParametersTest.java
trautmane: Yeah - I was being lazy.  I've split things up now.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/match/CopyMatchClient.java
trautmane: My intent/hope is that we flesh-out the validation implementations to do more later - assuming that is useful/helpful.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineParameters.java
trautmane: Yeah - I was going for explicit declaration in a small number of places rather than a pure generic solution.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/MFOVMontageMatchPatchClient.java
trautmane: For backwards compatibility, I was trying to keep support for running each client on its own outside of the new pipeline.  This is why there are createContextAndRun methods.  However, you are correct that when run as a pipeline, a single spark context will be shared and passed to the runPipelineStep method.

minnerbe: Thanks for the clarification! You are right, being able to run the clients on their own is desirable.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/pipeline/AlignmentPipelineStep.java
minnerbe: I wasn't familiar with that pattern, so I tried to come up with a "better" solution for several minutes, but I didn't succeed. So, I feel convinced that this should stay as it is. Your javadoc does help a lot, though!

minnerbe:
I had a look at the latest commits (and your comments), they look very good to me. From my side, I think we can go ahead and merge this.

PR #168, opened 2024-01-09
Restore old solver behavior

minnerbe:
I added a way of restoring the "old" way of alignment and intensity correction, based on the type of partitioning:
- z-partitioning: linearly blend the blocks, one solve is sufficient (old way).
- xy(z)-partitioning: pick random and solve multiple times alternatingly (new way).

The current design based on the `MergingStrategy` enum is somewhat ad hoc. However, I think that in order to introduce this feature more cleanly/naturally, other design issues have to be addressed first. Therefore, I would suggest to postpone implementing a "cleaner" solution to a later general refactoring step.

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/blockfactories/MergingStrategy.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/blockfactories/MergingStrategy.java
@@ -1,0 +1,6 @@
+package org.janelia.render.client.newsolver.blockfactories;
+
+public enum MergingStrategy {
+	RANDOM_PICK,
+	LINEAR_BLENDING;
+}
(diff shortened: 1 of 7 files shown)


render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/blockfactories/MergingStrategy.java
```diff
@@ -0,0 +1,6 @@
+package org.janelia.render.client.newsolver.blockfactories;
+
+public enum MergingStrategy {
+	RANDOM_PICK,
+	LINEAR_BLENDING;
+}
```
trautmane, line 6: consider adding a "list of models" supplier and moving the inetrpolateModels and pickRandom logic from the solver to here

trautmane (approved):
Looks good to me - one minor suggestion that you can take or leave as you like.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/blockfactories/MergingStrategy.java
minnerbe: I'm not sure I understand your comment. The logic in `interpolateModels` is not (yet) generic, but depends heavily on the model used in intensity correction and alignment. There already is a TODO for moving these two methods (and I agree that this enum is the canonical target) as soon as there is a proper model (not `List<...>`) for intensity correction.

trautmane: my mistake - thought logic could move here already - current todos are sufficient

PR #169, opened 2024-01-11
First attempt to add spark ZPositionCorrectionClient for use in

trautmane:
pipeline.  The pipeline implementation does not (yet) derive aligned stack names, so zcorr can only be run in isolation with the align stack names explicitly specified.

### DIFF
--- a/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
+++ b/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
@@ -162,7 +162,6 @@ public class ZPositionCorrectionClient
 
         final int completedSolveCount = rddCompletedSolveCount.collect().stream().reduce(0, Integer::sum);
 
-        LOG.info("deriveAndSolveCrossCorrelationData: collect RDD");
         LOG.info("deriveAndSolveCrossCorrelationData: exit, solved data for {} stacks", completedSolveCount);
     }
 
@@ -200,7 +199,6 @@ public class ZPositionCorrectionClient
 
         final int completedBatchCount = rddCompletedBatchCount.collect().stream().reduce(0, Integer::sum);
 
-        LOG.info("deriveCrossCorrelationData: collect RDD");
         LOG.info("deriveCrossCorrelationData: exit, generated data for {} batches", completedBatchCount);
     }
 
--- a/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
+++ b/render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
@@ -16,6 +16,7 @@ import org.janelia.alignment.spec.stack.StackWithZValues;
 import org.janelia.render.client.ClientRunner;
 import org.janelia.render.client.parameter.CommandLineParameters;
 import org.janelia.render.client.parameter.MultiProjectParameters;
+import org.janelia.render.client.parameter.ZRangeParameters;
 import org.janelia.render.client.parameter.ZSpacingParameters;
 import org.janelia.render.client.spark.LogUtilities;
 import org.janelia.render.client.spark.pipeline.AlignmentPipelineParameters;
@@ -46,6 +47,7 @@ public class ZPositionCorrectionClient
                                                                                             final boolean solveExisting)
                 throws IOException {
 
+            // TODO: think about alternative naming/packaging for related java and spark clients that would avoid need for long ugly declarations like this
             final org.janelia.render.client.zspacing.ZPositionCorrectionClient.Parameters jClientParameters =
                     new org.janelia.render.client.zspacing.ZPositionCorrectionClient.Parameters();
 
@@ -56,8 +58,8 @@ public class ZPositionCorrectionClient
             jClientParameters.renderWeb.project = stackId.getProject();
             jClientParameters.stack = stackId.getStack();
 
-            jClientParameters.layerRange.minZ = stackIdWithZValues.getFirstZ();
-            jClientParameters.layerRange.maxZ = stackIdWithZValues.getLastZ();
+            jClientParameters.layerRange = new ZRangeParameters(stackIdWithZValues.getFirstZ(),
+                                                                stackIdWithZValues.getLastZ());
 
             if (! solveExisting) {
                 // if deriving cross correlation for a batch of z layers (and not solving),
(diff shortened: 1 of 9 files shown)


render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
```diff
@@ -0,0 +1,206 @@
...
+        public ZSpacingParameters zSpacing = new ZSpacingParameters();
+
+        public org.janelia.render.client.zspacing.ZPositionCorrectionClient buildJavaClient(final StackWithZValues stackIdWithZValues,
+                                                                                            final int comparisonRange,
+                                                                                            final String runName,
+                                                                                            final boolean solveExisting)
+                throws IOException {
+
+            final org.janelia.render.client.zspacing.ZPositionCorrectionClient.Parameters jClientParameters =
+                    new org.janelia.render.client.zspacing.ZPositionCorrectionClient.Parameters();
```
minnerbe, line 50: That full name looks not very appealing. Is it possible to rename the spark client?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
```diff
@@ -0,0 +1,206 @@
...
+
+            jClientParameters.renderWeb.baseDataUrl = multiProject.baseDataUrl;
+
+            final StackId stackId = stackIdWithZValues.getStackId();
+            jClientParameters.renderWeb.owner = stackId.getOwner();
+            jClientParameters.renderWeb.project = stackId.getProject();
+            jClientParameters.stack = stackId.getStack();
+
+            jClientParameters.layerRange.minZ = stackIdWithZValues.getFirstZ();
+            jClientParameters.layerRange.maxZ = stackIdWithZValues.getLastZ();
```
minnerbe, line 60: Something about this "set public fields two layers in" doesn't seem right to me. Is there a way of supplying those parameters during construction of the parameter object?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
```diff
@@ -0,0 +1,206 @@
...
+        LOG.info("deriveAndSolveCrossCorrelationData: entry, clientParameters={}, comparisonRange={}",
+                 clientParameters, comparisonRange);
+
+        final MultiProjectParameters multiProjectParameters = clientParameters.multiProject;
+
+        final List<StackWithZValues> stackWithAllZValuesList = multiProjectParameters.buildListOfStackWithAllZ();
+
+        final long totalNumberOfZValues = stackWithAllZValuesList.stream()
+                .map(stackWithZ -> (long) stackWithZ.getzValues().size())
+                .reduce(0L, Long::sum);
```
minnerbe, line 139: Why are we using `long` here? Are we expecting more than 2 billion z-layers? Also, it seems that this is converted to `int` further down, anyway.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
```diff
@@ -0,0 +1,206 @@
...
+            final org.janelia.render.client.zspacing.ZPositionCorrectionClient jClient =
+                    clientParameters.buildJavaClient(stackWithZ, comparisonRange, runName, true);
+            final CrossCorrelationData ccData = jClient.loadCrossCorrelationDataSets();
+            jClient.estimateAndSaveZCoordinates(ccData);
+            return 1;
+        });
+
+        final int completedSolveCount = rddCompletedSolveCount.collect().stream().reduce(0, Integer::sum);
+
+        LOG.info("deriveAndSolveCrossCorrelationData: collect RDD");
```
minnerbe, line 163: I would expect this message before the actual collection step (otherwise I'd like it to read "collected RDD", in which case it is a bit redundant with the next message).

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
```diff
@@ -0,0 +1,206 @@
...
+            final org.janelia.render.client.zspacing.ZPositionCorrectionClient jClient =
+                    clientParameters.buildJavaClient(stackWithZ, comparisonRange, runName, false);
+            final CrossCorrelationData ccData = jClient.deriveCrossCorrelationData();
+            jClient.saveCrossCorrelationData(ccData);
+            return 1;
+        });
+
+        final int completedBatchCount = rddCompletedBatchCount.collect().stream().reduce(0, Integer::sum);
+
+        LOG.info("deriveCrossCorrelationData: collect RDD");
```
minnerbe, line 201: See above.

minnerbe (approved):
Apart from a few minor comments, this looks good.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
trautmane: The current pattern is to use the same base name with the `.spark` package name for the spark version of the clients.  This makes referencing the java clients from within the spark clients ugly.  I'll add a TODO for us to look into  other options as part of the upcoming overall client repackaging/renaming effort.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
trautmane: Yeah - I've always been a little conflicted about my choice to publicly expose the command line parameters.  Although I think this usage is consistent with the overall public setup, I went ahead and added a constructor as you requested.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
trautmane: I was just being super-careful since this supports multi-project / multi-stack runs.  Although it is very unlikely we'll ever need to process billions of z-layers, it didn't seem to hurt to use long here so that we don't have to worry about it.  The long is later converted to an int after batching (so after being divided by cluster parallelism) and will properly throw an exception if that conversion produces an overflow.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/zspacing/ZPositionCorrectionClient.java
trautmane: This is a bit of a hack.  The Spark driver logging used to write a dynamically updating progress bar that didn't finish with a carriage return.  I got in the habit of adding a superfluous log statement after each collect so that the next  log statement (the one I really wanted to see) would show up on its own new line.  It just checked recent logs and it looks like Spark no longer does this - or at least no longer does it by default.  So, I'll just remove the superfluous log statement.

PR #172, opened 2024-01-26
Document process for resuming alternating solve runs

trautmane:
To conserve compute resources, we would like to first (optimisitcally) solve with a smaller number of alternating runs (e.g. 2) and then have the option to resume the process with additional runs later if necessary.  I initially considered adding naming/suffix parameters to AlternatingRunParaemeters, but realized that things get messy trying to avoid naming conflicts with existing stacks since prior runs may or may not have kept intermediate stacks.  I decided it was easier to leave the current process as-is and handle resumes by specifying the previous aligned stacks as source stacks.  This will produce stacks with ugly aggregated names but it avoids conflicts and the stacks can be renamed later with a script if desired.

To resume a run, you specify the previous aligned result as a "raw" source and set shiftBlocks to true if there were an odd number of prior runs:
```
  "pipelineStackGroups": {
    "raw": {
      "projectPattern": "^cut_000_to_009$",
      "stackPattern": "^c..._s..._v01_align$"   // instead of "^c..._s..._v01$"
    },
    ...
  },
  "pipelineSteps": [
    "ALIGN_TILES"
  ],
  ...
  "affineBlockSolverSetup": {
    ...
    "blockPartition": {
      ...
      "shiftBlocks": false                      // set to true if odd number of prior runs
    },
    "alternatingRuns": {
      "nRuns": 2,
      "keepIntermediateStacks": false
    }
  },

```

Here are what the stack names might look like when resuming a run that had 2 prior runs:
```
first run:
  source: c009_s310_v01 ->
    c009_s310_v01_align_run1
    c009_s310_v01_align_run2 (or c009_s310_v01_align if keepIntermediateStacks false)

resumed run:
  source: c009_s310_v01_align_run2 ->
    c009_s310_v01_align_run2_align_run1
    c009_s310_v01_align_run2_align_run2 (or c009_s310_v01_align_run2_align if keepIntermediateStacks false)
```

Let me know if you think this approach is reasonable or if you think there is a better way.

Although I did not end up adding any code to support resuming alternating runs, I did improve the spark distribute affine solve parallelization and fixed a couple of bugs so there is a little related code to review :).

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
@@ -136,7 +136,7 @@ public class StackIdWithZParameters
         return batchedList;
     }
 
-    private List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
+    public List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
             throws IOException {
 
         final StackIdNamingGroup defaultGroup = new StackIdNamingGroup(projectPattern, stackPattern);
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
@@ -136,7 +136,7 @@ public class StackIdWithZParameters
         return batchedList;
     }
 
-    public List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
+    private List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
             throws IOException {
 
         final StackIdNamingGroup defaultGroup = new StackIdNamingGroup(projectPattern, stackPattern);
--- a/render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
+++ b/render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
@@ -40,34 +40,6 @@ public class StackIdWithZParametersTest {
     private final RenderDataClient mockDataClient = new MockRenderDataClient(owner, projectA, stackIds);
 
     @Test
-    public void testGetEligibleStackIds() throws IOException {
-
-        final StackIdWithZParameters params = new StackIdWithZParameters();
-
-        params.projectPattern = ".*ProjectA$";
-        List<StackId> eligibleStackIds = params.getEligibleStackIds(mockDataClient);
-        Assert.assertEquals("all stacks should be returned with default project pattern",
-                            12, eligibleStackIds.size());
-
-        params.projectPattern = null;
-        params.stackPattern = "stack1.*";
-        eligibleStackIds = params.getEligibleStackIds(mockDataClient);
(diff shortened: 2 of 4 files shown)


render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
```diff
@@ -136,7 +136,7 @@ public List<StackWithZValues> buildListOfStackWithBatchedZ(final RenderDataClien
         return batchedList;
     }
 
-    private List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
+    public List<StackId> getEligibleStackIds(final RenderDataClient renderDataClient)
```
minnerbe, line 139: I know I've been a lot more neglectful than you with commenting in the past, but I suggest we add a doc string if some method's visibility is set to public.

render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
```diff
@@ -0,0 +1,127 @@
...
+            new StackId(owner, projectB, "stack1_align"),
+            new StackId(owner, projectB, "stack2_align"),
+            new StackId(owner, projectB, "stack1_align_ic"),
+            new StackId(owner, projectB, "stack2_align_ic")
+    );
+
+    private final RenderDataClient mockDataClient = new MockRenderDataClient(owner, projectA, stackIds);
+
+    @Test
+    public void testGetEligibleStackIds() throws IOException {
```
minnerbe, line 43: It seems to me that this doesn't really test the behavior of `StackIdWithZParameters`, but just tests if the line
```Java
return hasProjectPattern ? renderDataClient.getOwnerStacks() : renderDataClient.getProjectStacks();
```
does the right thing (which I would classify as implementation detail and most of which is mocked anyway).

Was `getEligibleStackIds` only made public to be able to test it here? If so, it should probably remain private and not be tested at all.

minnerbe (changes requested):
Except for the comment in the test, everything else looks good to me. I would suggest to simply revert the visibility of `getEligibleStackIds` to `private` and not test it. Since it is part of the observable behavior, this method is implicitly tested anyway.

To your question: the only alternative solution that comes to my mind is to explicitly introduce a parameter `int resumeFromRun` to mark the given stack as "run n" from a previous execution. Then, we could revert appending the "_align" suffix and just go on with the regular naming scheme. This is probably some work, though, and I don't see how this is better than your current solution (except for the minor inconvenience of "_align_runN" stacks). So, I would just leave it as it is for now and see how often we actually need to do extra runs.

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/StackIdWithZParameters.java
trautmane: agreed - I think it is good practice to javadoc anything public

render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
trautmane: Yeah - I drafted this originally because I hadn't found the flipped logic bug in StackIdNamingGroup yet.  I agree it is easier to drop the test and restore getEligibleStackIds to a private method (and rely on testGetStackIdList to catch future issues).

PR #174, opened 2024-02-01
Make intensity correction faster by downsampling cross layer matches more

minnerbe:
I added a parameter to the intensity matching process that allows to specify a different render scale for cross layer matches than for same layer matches. So far, the bottleneck were matches between tiles in the same position but on different layers, since they have a significantly larger overlap (the time was spent in outlier detection since there were a lot of matches).

### Details
This might seem like a big change, but the only novelty was [introducing and handling the additional scale parameter](https://github.com/saalfeldlab/render/commit/e870102af27c82f5273aa2e5877a92564ee49e10). All other changes were made while cleaning up. In particular, I:
- extracted the formerly nested `Matcher` class into its own file and moved the average computation there as well;
- removed unnecessary ImgLib2 calls (some `RealInterval...` for the bounding box of a tile and `ListImage` for the matching process).

For further clean up, we should probably think about making the actual rendering process into a class with state instead of a static `Render.render` method.

### Verification
I tested the new parameter with 2 layers of c009_s310_v01_mfov_08 and `--crossLayerRenderScale 0.02` (as opposed to the default scale `0.1`). [Here](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B45306.85546875%2C7255.81005859375%2C440%5D%2C%22crossSectionScale%22:8.780986177504422%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafer_53/cut_000_to_009/c009_s310_v01_mfov_08_ic_test%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22c009_s310_v01_mfov_08_ic_test%22%7D%5D%2C%22selectedLayer%22:%7B%22layer%22:%22c009_s310_v01_mfov_08_ic_test%22%7D%2C%22layout%22:%22xy%22%7D) is a neuroglancer view of the two layers, which look good to me.

### Improvement
At scale `0.1`, matching two tiles that partially overlap takes about 150-300ms, while matching tiles that completely overlap takes roughly 1900ms. At scale `0.02`, matching tiles that completely overlap only takes about 200-300ms.
This reduced the runtime from 9:38min (where matching took 6:45min) to 5:00min (where matching took 3:10min). The additional gain in runtime probably comes from the fact that there are fewer matching tile pairs, since the overlap of some tiles across layers is too small to produce matches at scale `0.02`.

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
@@ -6,10 +6,9 @@ import mpicbg.models.Affine1D;
 import mpicbg.models.Point;
 import mpicbg.models.PointMatch;
 import mpicbg.models.Tile;
-import net.imglib2.Interval;
 import net.imglib2.img.list.ListImg;
 import net.imglib2.img.list.ListRandomAccess;
-import net.imglib2.util.Intervals;
+import net.imglib2.util.Pair;
 import net.imglib2.util.StopWatch;
 import net.imglib2.util.ValuePair;
 import org.janelia.alignment.spec.TileSpec;
@@ -23,10 +22,7 @@ import java.awt.*;
 import java.util.ArrayList;
 import java.util.HashMap;
 
-class IntensityMatcher implements Runnable {
-	//final private Rectangle roi;
-	final private ValuePair<TileSpec, TileSpec> patchPair;
-	final private HashMap<String, ArrayList<Tile<? extends Affine1D<?>>>> coefficientTiles;
+class IntensityMatcher {
 	final private PointMatchFilter filter;
 	final private double scale;
 	final private int numCoefficients;
@@ -34,15 +30,11 @@ class IntensityMatcher implements Runnable {
 	final ImageProcessorCache imageProcessorCache;
 
 	public IntensityMatcher(
-			final ValuePair<TileSpec, TileSpec> patchPair,
-			final HashMap<String, ArrayList<Tile<? extends Affine1D<?>>>> coefficientTiles,
 			final PointMatchFilter filter,
 			final double scale,
 			final int numCoefficients,
 			final int meshResolution,
 			final ImageProcessorCache imageProcessorCache) {
-		this.patchPair = patchPair;
-		this.coefficientTiles = coefficientTiles;
 		this.filter = filter;
 		this.scale = scale;
 		this.numCoefficients = numCoefficients;
@@ -50,13 +42,9 @@ class IntensityMatcher implements Runnable {
 		this.imageProcessorCache = imageProcessorCache;
 	}
 
-	@Override
-	public void run() {
-		final TileSpec p1 = patchPair.getA();
-		final TileSpec p2 = patchPair.getB();
+	public void match(final TileSpec p1, final TileSpec p2, final HashMap<String, ArrayList<Tile<? extends Affine1D<?>>>> coefficientTiles) {
 
 		final StopWatch stopWatch = StopWatch.createAndStart();
-
 		LOG.info("run: entry, pair {} <-> {}", p1.getTileId(), p2.getTileId());
 
 		final Rectangle box = computeIntersection(p1, p2);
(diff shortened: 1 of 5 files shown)


render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
```diff
@@ -0,0 +1,192 @@
...
+		this.crossLayerScale = parameters.crossLayerRenderScale();
+		this.numCoefficients = parameters.numCoefficients();
+		this.meshResolution = meshResolution;
+		this.imageProcessorCache = imageProcessorCache;
+	}
+
+	public void match(final TileSpec p1, final TileSpec p2, final HashMap<String, ArrayList<Tile<? extends Affine1D<?>>>> coefficientTiles) {
+
+		final StopWatch stopWatch = StopWatch.createAndStart();
+		LOG.info("match: entry, pair {} <-> {}", p1.getTileId(), p2.getTileId());
```
trautmane, line 47: The logging is pretty verbose here (even for me :).  Since matching typically takes at most a couple of seconds, I suggest removing all of the `LOG.info` calls from this method other than the last one at exit which provides all of the info we really need.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
```diff
@@ -0,0 +1,192 @@
...
+		final FloatProcessor weights1 = new FloatProcessor(w, h);
+		final ColorProcessor subTiles1 = new ColorProcessor(w, h);
+		final FloatProcessor pixels2 = new FloatProcessor(w, h);
+		final FloatProcessor weights2 = new FloatProcessor(w, h);
+		final ColorProcessor subTiles2 = new ColorProcessor(w, h);
+
+		Render.render(p1, numCoefficients, numCoefficients, pixels1, weights1, subTiles1, box.x, box.y, scale, meshResolution, imageProcessorCache);
+		Render.render(p2, numCoefficients, numCoefficients, pixels2, weights2, subTiles2, box.x, box.y, scale, meshResolution, imageProcessorCache);
+
+		LOG.info("match: generate matrix for pair {} <-> {} and filter", p1.getTileId(), p2.getTileId());
```
trautmane, line 65: see above

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
```diff
@@ -0,0 +1,192 @@
...
+		return matrix.get(i * size + j);
+	}
+
+	private static int numberOfPixels(final int length, final double scale) {
+		return (int) Math.round(length * scale);
+	}
+
+	List<Double> computeAverages(final TileSpec tile) {
+
+		LOG.info("computeAverages: entry, tile {}", tile.getTileId());
```
trautmane, line 144: probably don't need this log statement since exit statement includes time stats

trautmane (approved):
looks good - made some minor comments as usual

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/AlgorithmicIntensityAdjustParameters.java
```diff
@@ -34,15 +34,20 @@ public class AlgorithmicIntensityAdjustParameters implements Serializable {
...
 			names = "--renderScale",
-			description = "Scale for rendered tiles used during intensity comparison")
+			description = "Scale for rendered tiles in the same layer used during intensity comparison (default: 0.1)")
 	public double renderScale = 0.1;
 
+	@Parameter(
+			names = "--crossLayerRenderScale",
+			description = "Scale for rendered tiles in different layers used during intensity comparison (default: same as renderScale)")
+	public double crossLayerRenderScale = 0.0;
+
```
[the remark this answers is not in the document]
minnerbe: I disagree that we disagree here. Thanks for the suggestion!

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/IntensityMatcher.java
```diff
@@ -0,0 +1,192 @@
...
+import org.janelia.render.client.newsolver.blocksolveparameters.FIBSEMIntensityCorrectionParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.awt.Rectangle;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+class IntensityMatcher {
```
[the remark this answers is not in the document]
minnerbe: I'll look into that, but there's a good chance that the other class is a remnant of the "old solver". In that case, I would just leave it there and delete it when getting rid of the old solver files.

PR #176, opened 2024-02-05
Sparkify intensity correction

minnerbe:
This is my first attempt at a spark client for intensity correction. Since this is my first attempt at spark ever, I copied a lot of stuff and I'm sure that there are a bunch of bugs in there.
I'm happy to go over this together at some point to learn from your experience @trautmane. 

One thing that I take from this experience is that we didn't do a good job of abstracting concepts. I tried to extract common functionality but failed for various reasons most of the time:
* `AffineBlockSolverSetup` and `IntensityCorrectionSetup` don't share a common superclass or interface.
* The "model" for intensity correction does not implement `Model`, which makes it hard to generalize.
* There are a lot of `public static` methods. While this is not bad per se, I think here it means that encapsulation is not tight.

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/setup/IntensityCorrectionSetup.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/setup/IntensityCorrectionSetup.java
@@ -3,6 +3,9 @@ package org.janelia.render.client.newsolver.setup;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.beust.jcommander.ParametersDelegate;
+import org.janelia.alignment.json.JsonUtils;
+import org.janelia.alignment.spec.stack.StackId;
+import org.janelia.alignment.spec.stack.StackWithZValues;
 import org.janelia.render.client.parameter.AlgorithmicIntensityAdjustParameters;
 import org.janelia.render.client.parameter.AlternatingRunParameters;
 import org.janelia.render.client.parameter.CommandLineParameters;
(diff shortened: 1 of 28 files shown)


render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/setup/IntensityCorrectionSetup.java
```diff
@@ -58,4 +61,52 @@ public void initDefaultValues() {
...
+	 *
+	 * @param  matchSuffix                            suffix to append to derived match collection names
+	 *                                                (specify empty string to omit suffix).
+	 *                                                Suffix is needed when match aggregation is performed
+	 *                                                by an earlier pipeline step.
+	 * @return a clone of this setup populated with the specified parameters.
+	 */
+	public IntensityCorrectionSetup buildPipelineClone(final String baseDataUrl,
+													   final StackWithZValues stackWithZValues,
+													   final boolean deriveMatchCollectionNamesFromProject,
```
trautmane: I removed unused match parameters in a commit below.

trautmane (approved):
Looks good to me at first glance.  I agree that there are opportunities for improved abstraction later.  I've made a few commits myself - let's discuss/review those together.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/newsolver/DistributedAlternatingDomainDecompositionSolver.java
```diff
@@ -0,0 +1,85 @@
...
+import org.janelia.render.client.RenderDataClient;
+import org.janelia.render.client.newsolver.setup.DistributedSolveParameters;
+import org.janelia.render.client.parameter.RenderWebServiceParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+
+public abstract class DistributedAlternatingDomainDecompositionSolver {
```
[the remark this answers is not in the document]
minnerbe: I agree. As I said, my initial intent was to pull more of the common functionality up, but this wasn't possible for reasons that are out of scope for this PR.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/setup/IntensityCorrectionSetup.java
minnerbe: Thanks! I completely overlooked the mention of matches, here.

PR #177, opened 2024-02-06
Deploy render-ws within Docker swarm

trautmane:
First draft of scripts/configuration to deploy render-ws on docker swarm with multiple nodes.  Previous approach at Janelia was to deploy using docker compose on a single node.  Intent is to experiment with new deployment to see if it helps with the concurrent request failures observed during large scale multi-SEM alignment pipeline runs.


render-ws/src/main/scripts/docker-compose/02_docker_swarm.sh
```diff
@@ -0,0 +1,124 @@
...
+  exit 1
+fi
+
+if (( $# == 2 )); then
+  export RENDER_WS_IMAGE="${2}"
+else
+  export RENDER_WS_IMAGE="${2}:${3}"
+fi
+
+if [ "${JETTY_RUN_AS_USER_AND_GROUP_IDS}" == "999:999" ]; then
```
minnerbe, line 74: It's not clear to me where `JETTY_RUN_AS_USER_AND_GROUP_IDS` would be changed from the default value to prevent this error from being thrown.

render-ws/src/main/scripts/docker-compose/02_docker_swarm.sh
```diff
@@ -0,0 +1,124 @@
...
+$(date) running:
+  docker stack up --compose-file ${DOCKER_COMPOSE_YML} --prune --resolve-image always ${SWARM_STACK_NAME}
+"
+
+echo "${LAUNCH_INFO}" >> "${SWARM_OUT_FILE}"
+
+nohup docker stack up --compose-file "${DOCKER_COMPOSE_YML}" --prune --resolve-image always "${SWARM_STACK_NAME}" >> "${SWARM_OUT_FILE}" 2>&1 &
+sleep 5
+
+docker node ls | tee -a "${SWARM_OUT_FILE}"
```
minnerbe, line 124: I see a red circled minus, here. Took me a while to figure out that this means "no newline at the end of the file". What's our policy on this?

render-ws/src/main/scripts/docker-compose/docker-compose.swarm.yml
```diff
@@ -0,0 +1,54 @@
...
+      - "8080:8080"
+
+    # Specify configuration related to the deployment and running of services.
+    deploy:
+
+      # One instance of the service on every node in the swarm.
+      mode: global
+
+      # Docker assigns the service a virtual IP (VIP) that acts as the front end for clients to reach the service on a network.
+      # Docker ingress routing mesh routes requests between the client and available worker nodes for the service,
```
minnerbe, line 16: It took me a while to parse that sentence, because I wasn't familiar with the term "Docker ingress routing mesh". Is there a way of making this comment clearer, or is it clear to people familiar with Docker, anyway?

minnerbe (approved):
As I'm not too familiar with Docker, I cannot really give valuable feedback for this PR.
However, I made some comments where things were not immediately clear to me. Those are questions for my understanding rather than points of critique, so feel free to ignore them.

render-ws/src/main/scripts/docker-compose/02_docker_swarm.sh
trautmane: The script is intended to by copied and (minimally) edited for each deployment environment.  This variable should be changed on line 13.  I'll improve the error message.

PR #178, opened 2024-02-17
Make (i)FFT use correct number of threads

minnerbe:
While skimming through the streak-correction code, I realized that the custom number of threads is set *after* processing the (inverse) FFT. This seems incorrect to me.

Can you confirm this, @StephanPreibisch ?


StephanPreibisch:
Looks good to me, little bug ... thanks for catching it!

StephanPreibisch (commented):
looks good!

PR #179, opened 2024-02-22
Feature: local destreaking

minnerbe:
I implemented a first version of local de-streaking. The locality is based on the following algorithm:
1. de-streak with band-pass
2. threshold difference to get only the streaks
3. blur the result and threshold again to get a larger area around the streaks
4. blur the result again to get a smooth transition
5. use the result as a local weight between the original and the de-streaked image

The locality works reasonably well, the overall result looks not that convincing, though. This may be based on the non-linear intensity transformation that the de-streaking introduces (there are lighter and darker regions simultaneously).

Side note: My manual tests in FIJI looked a little better, which could be due to the different (i.e., not as invasive) band-pass that I applied there.

Screenshot from 2024-02-22 18-17-08

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/destreak/ConfigurableMaskStreakCorrector.java
+++ b/render-app/src/main/java/org/janelia/alignment/destreak/ConfigurableMaskStreakCorrector.java
@@ -2,24 +2,20 @@ package org.janelia.alignment.destreak;
 
 import ij.ImagePlus;
 import ij.process.ImageProcessor;
-
-import java.util.HashMap;
-import java.util.Map;
-
-import org.janelia.alignment.filter.Filter;
-import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


minnerbe:
I implemented `SmoothMaskStreakCorrector`, a new smoother version of the FFT based band-pass filter that we had (see results below). IMHO, this yields much better results and in particular doesn't change local intensity that drastically.

Also, the new method has 3 tunable parameters:
* distance of cutoff from the middle
* width of cutoff band
* angle of cutoff band (orthogonal to streaks)

This should be a lot easier to tune than the current manual workflow.
Screenshot from 2024-02-23 14-08-58

minnerbe:
Another update: a local version of the smoother mask works.

Screenshot from 2024-02-23 17-28-49

PR #180, opened 2024-03-03
Fix: streak correction

minnerbe:
I tried to make the streak correction code a bit more robust:
* All computations in non-local `StreakCorrector`s are done in `FloatType`
* All computations in local ones are done in `UnsignedByteType`
* All wrappings have a check that blows up with a sensible error message if the wrapping didn't work


trautmane (approved):
looks good to me

PR #181, opened 2024-03-12
Add residuals option to StackAlignmentErrorClient

trautmane:


trautmane:
Added a residuals option to the StackAlignmentErrorClient.  Fixed stackInGroup logic bug.  Hacked around a bug in the shifted z when blockSize for z is 1 - let's discuss this when you return.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
```diff
@@ -37,6 +47,11 @@ public static class Parameters extends CommandLineParameters {
 		private String stack;
 		@Parameter(names = "--fileName", description = "Name of file to write pairwise errors to (default: errors_<stack>.json.gz)")
 		private String fileName = null;
+		@Parameter(
+				names = "--calculateResiduals",
+				description = "Calculate root mean squared error for each pair of tiles (default: calculate Preibisch error)",
```
minnerbe, line 52: We should probably rename the errors to something more expressive at some point. In the long run, it would also be cool to handle the different error measures in a polymorphic way, instead of boolean flags. Nothing to be taken care of right now, but a todo probably wouldn't hurt.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
```diff
@@ -122,27 +158,95 @@ private static AlignmentErrors computeSolveItemErrors(final ResolvedTileSpecsWit
...
+		final ResidualCalculator.Result result = residualCalculator.run(stackId,
+																		inputData,
+																		localMatchList,
+																		pAlignedTileSpec,
+																		qAlignedTileSpec);
+
+		return result.getRootMeanSquareError();
+	}
+
+	private static TileSpec getMatchTileSpec(final TileSpec alignedTileSpec,
```
minnerbe, line 228: Would `getUntransformedTileSpec` be a better name for this method?
Also, it could be precomputed for all nodes in `tilesAndMatches.getTiles()` before looping over all matches. This would probably be more reusable and the way it's done right now introduces some mental clutter in form of a mutable map.

render-ws-java-client/src/test/java/org/janelia/render/client/BlockLayoutCreatorTest.java
```diff
@@ -134,4 +133,40 @@ public void specifyingDimensionTwiceThrowsError() {
 			assertEquals("Intervals for dimension x already specified", e.getMessage());
 		}
 	}
+
+	@Test
+	public void sameMaxZForRegularAndShiftedGrids() {
```
minnerbe, line 138: I think this should be renamed or split into two tests:
* the regular and shifted parts are independent,
* both could fail by having the same `maxZ` that is different from the one of `bounds`, which would not contradict the test name.

render-ws-java-client/src/test/java/org/janelia/render/client/BlockLayoutCreatorTest.java
```diff
@@ -134,4 +133,40 @@ public void specifyingDimensionTwiceThrowsError() {
...
+
+		final int blockSizeXY = 9000;
+		final int blockSizeZ = 1;
+		final Bounds bounds = new Bounds(73.0, 164.0, 11986.0,
```
minnerbe, line 142: Where do the numbers come from? Is this a bug that occurred in practice or just random numbers? In the first case, I would add a short comment. In the second case, I'm wondering if there was a less random set of numbers that show the same properties of the system under test (maybe in a more straightforward way)?

render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
```diff
@@ -66,6 +66,12 @@ public void testGetStackIdList() throws IOException {
                 .filter(stackId -> stackId.getProject().equals(projectB)).count();
         Assert.assertEquals("all stacks should be from project B (naming group should override default)",
                             6, projectBStackCount);
+
```
minnerbe, line 69: `testGetStackIdList()` should probably also be split in multiple tests, but this is not related to the PR under review.

minnerbe (approved):
The logic and the fixes look good to me. I have a few questions, mainly because I didn't have time to think too deeply about them -- so feel free to ignore.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
trautmane: makes sense - I'll add a todo

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
trautmane: With regard  to the method name ... I don't think `getUntransformedTileSpec` is accurate because the tile spec may still be transformed for lens correction.  I'll switch this to `buildTileSpecUsedForMatchDerivation` and hope that is clearer.  As you suggested, I now build the tileIdToMatchTileSpec map up front rather than as we go.

render-ws-java-client/src/test/java/org/janelia/render/client/BlockLayoutCreatorTest.java
trautmane: sure - I split it into 2 tests

render-ws-java-client/src/test/java/org/janelia/render/client/BlockLayoutCreatorTest.java
trautmane: This is a bug that occurred in practice.  I added a comment to explain.

render-ws-java-client/src/test/java/org/janelia/render/client/parameter/StackIdWithZParametersTest.java
trautmane: Yeah - I'm ok with it either way.  If you feel compelled to split it up, that's fine - I'll leave it to you to do :).

PR #186, opened 2024-04-02
Neuroglancer links for worst errors

minnerbe:
I refactored the client that computes an error measure on every (same layer) pair of tiles in an aligned stack:
- it does not do comparison by default anymore, but a second stack to act as a baseline is optional;
- it does not write all the errors to disc anymore;
- instead, it logs the N worst pairs and assembles a neuroglancer link that shows exactly that pair.


render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
```diff
@@ -2,85 +2,77 @@
...
-		// TODO: define enum for different error types (Preibisch, RMSE, ...) and use that instead of boolean flag
-
-		public String getFileName() {
-			if (fileName == null) {
-				fileName = "errors_" + stack + ".json.gz";
-			}
-			return fileName;
-		}
+				names = "--errorMetric",
+				description = "Error metric to use for computing errors (default: GLOBAL_LOCAL_DIFFERENCE)")
```
trautmane, line 57: There is no need to put default values in the description because utility automatically does that.

trautmane (approved):
looks good to me - just a minor comment on leaving out default values from parameter description text

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/StackAlignmentErrorClient.java
minnerbe: Oh, thanks for clarifying! I didn't know (and apparently never noticed). I'll remove default value description in the future if they can be inferred from context.

PR #187, opened 2024-04-09
Reordering tiles

minnerbe:
I investigated the rendering order for tiles in multi-SEM and implemented a quick hack to re-order tiles to get the desired rendering order: for each tile in a z-layer, a new serial number is created (that reflects the desired rendering order) and prepended to the tile ID.

There are currently three orders implemented:
1. the original order (i.e., don't change anything)
2. render the mFOVs last to first and the sFOVs in a left to right, top to bottom fashion
3. the reverse of 2.

It looks like strategy 2 yields the best results as most tiles have the unfavorable regions on the right and the bottom of the image. However, some tiles have one of the regions on the left side instead of the right side, or it is completely missing. In theory, one should be able to determine the location (and therefore the correct order) by investigating the y-coordinates of the tiles. However, I didn't find any definitive criterion (based on bounding boxes before or after transformation) that could solve this. Therefore, I suggest to stay with strategy 2 as the default for now.

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/TileReorderingClient.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/TileReorderingClient.java
@@ -24,27 +24,6 @@ import java.util.stream.Collectors;
  */
 public class TileReorderingClient {
 
-	private static final int SFOV_INDEX = 2;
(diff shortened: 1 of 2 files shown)


trautmane (approved):
looks good - I added one minor comment to clarify match limitations

PR #188, opened 2024-04-23
Flatfield visualization and application

minnerbe:
I added some code to visualize flat field estimates and apply them to the wafer_53 images.

There are two new clients:
* `FlatfieldVisualizationClient` creates a stack that shows all flat field estimates on a regular hexagonal grid (there is one flat field estimate per z-layer and per sfov, making one mfov column in total);
* `MultiSemFlatFieldCorrectionClient` re-saves all source images of a collection of stacks and creates new stacks that point to the re-saved images.

For now, the flat field estimates are created externally, but the plan is to port this to render at a later time.


render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+		private String flatFieldLocation;
+		@Parameter(names = "--outputRoot", description = "Folder to write corrected images to", required = true)
+		private String outputFolder;
+		@Parameter(names = "--flatFieldFormat", description = "Format of the flat field estimates in printf style with placeholders for z-layer and sfov", required = true)
+		private String flatFieldFormat;
+		@Parameter(names = "--targetStackSuffix", description = "Suffix to append to the stack name for the corrected stack", required = true)
+		private String targetStackSuffix = "_corrected";
+		@Parameter(names = "--inputRoot", description = "Root folder for input data; if given, the structure under this root is replicated in the output folder")
+		private String inputRoot = null;
+		@Parameter(names = "--flatFieldConstantFromZ", description = "Maximum z-layer of flat field estimates to consider. All subsequent z-layers get corrected with the maxium, since the estimates can be bad for the last few z-layers. If not given, all z-layers are considered")
```
trautmane, line 54: Should this be something like `--finalZLayersToIgnore` instead with a default value of zero instead since the multi-SEM stacks have different total z layer counts?

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+		};
+		clientRunner.run();
+	}
+
+	public MultiSemFlatFieldCorrectionClient(final Parameters parameters) {
+		this.params = parameters;
+		this.renderClient = new RenderDataClient(parameters.multiProject.baseDataUrl, parameters.multiProject.owner, parameters.multiProject.project);
+	}
+
+	public void correctTiles() throws IOException {
```
trautmane, line 79: Though not sure this is needed, to facilitate future usage within our spark pipeline setup, I'd separate this into 2 methods, pass the stack info in, and construct the render data client locally ... something like:

```
    public void correctStacks(final String baseDataUrl,
                              final List<StackId> stackIds,
                              final String targetStackSuffix) throws IOException {
        for (final StackId stackId : stackIds) {
            final String targetStackName = stackId.getStack() + targetStackSuffix;
            correctTilesForStack(baseDataUrl, stackId, targetStackName);
        }
    }

    public void correctTilesForStack(final String baseDataUrl,
                                     final StackId sourceStackId,
                                     final String targetStackName) throws IOException {

        final RenderDataClient renderClient = new RenderDataClient(baseDataUrl, sourceStackId.getOwner(), sourceStackId.getProject());

        final String sourceStackName = sourceStackId.getStack();
        final StackMetaData sourceStackMetaData = renderClient.getStackMetaData(sourceStackName);

        renderClient.setupDerivedStack(sourceStackMetaData, targetStackName);

        for (final double z : renderClient.getStackZValues(sourceStackId.getStack())) {
            ...

```

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+					saveImage(ip, tileSpec);
+				}
+
+				renderClient.saveResolvedTiles(tileSpecs, stack.getStack() + params.targetStackSuffix, (double) z);
+			}
+			renderClient.setStackState(stack.getStack() + params.targetStackSuffix, StackMetaData.StackState.COMPLETE);
+		}
+	}
+
+	private DoubleSummaryStatistics getZStatistics(final StackId stack) throws IOException {
```
trautmane, line 108: I didn't see the need for this

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Apply a previously computed flat field estimate to correct shading in a Multi-SEM project.
+ * Originally developed for the wafer_53 Multi-SEM dataset.
+ *
+ * @author Michael Innerberger
+ */
+public class MultiSemFlatFieldCorrectionClient {
+
+	// make cache large enough to hold all flat field estimates for one layer
+	private static final ImageProcessorCache CACHE = new ImageProcessorCache(91L * 2000L * 1748L + 1, false, false);
```
trautmane, line 36: This should eventually be parameterized

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	private ImageProcessor loadFlatFieldEstimate(final int z, final int sfov) {
+		final Path imagePath = Path.of(params.flatFieldLocation, String.format(params.flatFieldFormat, z, sfov));
+		final String imageUrl = "file:" + imagePath;
+		final ImageLoader.LoaderType loaderType = ImageLoader.LoaderType.IMAGEJ_DEFAULT;
+		return CACHE.get(imageUrl, 0, false, false, loaderType, null);
+	}
+
+	private int extractSfovNumber(final TileSpec tileSpec) {
```
trautmane, line 122: This is also implemented here: https://github.com/saalfeldlab/render/blob/newsolver/render-ws-java-client/src/main/java/org/janelia/render/client/multisem/Utilities.java#L61

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
```diff
@@ -0,0 +1,195 @@
...
+		final Path newPath;
+
+		if (params.inputRoot != null) {
+			final Path relativePath = Path.of(params.inputRoot).relativize(originalPath);
+			newPath = Path.of(params.outputFolder).resolve(relativePath);
+		} else {
+			newPath = Path.of(params.outputFolder).resolve(originalPath.getFileName());
+		}
+
+		ensureFolderExists(newPath.getParent());
```
trautmane, line 159: I would separate file system stuff from tile spec manipulation and ensure the folder exists as part of the saveImage process.

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
minnerbe: The idea is that the flat field estimates for the last couple of z-layers are bad, because there's only very little data. Thus, Preibisch suggested to just take the last z-layer which has a decent estimate for all subsequent layers (i.e., keep the flat field constant from a particular z-layer, hence the name).
I struggled to find a good name for the parameter and a concise description. If you have a better suggestion, I'm very open to that.

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
minnerbe: That's good to know, thanks! This also applies to the other PR for the re-ordering.

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
minnerbe: That is a remainder from when I wanted to compute the min and max for all the stacks up front. You are right, this is not needed anymore and I'll refactor this to just use the stack's own z-values.

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
trautmane: Makes sense - my guess was that you don't always want say z 35, but instead want maxZ - 4 because maxZ can range from 3? to 4?

render-ws-java-client/src/main/java/org/janelia/render/client/MultiSemFlatFieldCorrectionClient.java
minnerbe: The provided functionality is not the same: I really need the number of the sfov, but the given utility class also returns the scan and mfov numbers.

PR #189, opened 2024-05-01
Exponential fit

minnerbe:
I added an exponential fit that could potentially cure the dark area at the top of most tiles. To this end, I introduced a client that estimates the correction and adds a filter spec to the tile specs, as well as a filter.

There is [one (layer of one) stack](http://renderer.int.janelia.org:8080/render-ws/view/point-match-explorer.html?renderDataHost=renderer.int.janelia.org%3A8080&dynamicRenderHost=renderer.int.janelia.org%3A8080&catmaidHost=renderer-catmaid.int.janelia.org%3A8000&ndvizHost=renderer.int.janelia.org%3A8080&renderStackOwner=hess_wafer_53d&renderStackProject=slab_070_to_079&renderStack=s070_m104_align_exp_test02&startZ=2&endZ=2) that already has this correction included. However, since the filter code is not deployed yet, I could check the quality of the correction only locally so far.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/filter/CompositeFilter.java
+++ b/render-app/src/main/java/org/janelia/alignment/filter/CompositeFilter.java
@@ -1,0 +1,58 @@
+package org.janelia.alignment.filter;
+
+import ij.process.ImageProcessor;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Holds a collection of {@link Filter}s to be applied in sequence.
+ *
+ * @author Michael Innerberger
+ */
+public class CompositeFilter implements Filter {
+
+    private List<Filter> filters;
+
+    // empty constructor required to create instances from specifications
+    @SuppressWarnings("unused")
+    public CompositeFilter() {
+        this((List<Filter>) null);
+    }
+
+    public CompositeFilter(final Filter... filters) {
+        this(List.of(filters));
+    }
+
+    public CompositeFilter(final List<Filter> filters) {
+        this.filters = filters;
+    }
+
+    @Override
+    public void init(final Map<String, String> params) {
+        filters = new ArrayList<>(params.size());
+        for (final String serializedFilter : params.values()) {
+            final FilterSpec filterSpec = FilterSpec.fromJson(serializedFilter);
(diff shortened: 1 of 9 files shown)


render-ws-java-client/src/main/java/org/janelia/render/client/multisem/ExponentialFitClient.java
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	private PrintWriter getWriterIfDesired() throws FileNotFoundException {
+		if (params.coefficientsFile == null) {
+			return null;
+		}
+		final File coefficientsFile = new File(params.coefficientsFile);
+		if (coefficientsFile.exists()) {
+			LOG.error("process: coefficients file {} already exists", params.coefficientsFile);
+			System.exit(1);
```
trautmane, line 131: I would just throw an IllegalArgumentException (or IOException if that reduces throws statements) here rather than calling System.exit.

minnerbe: You're right, thanks! I still tend to think in terms of local execution.

minnerbe:
@trautmane I pushed some updates, could you please also review these? Thanks!

trautmane (commented):
nice - one concern about filter ordering ...

render-app/src/main/java/org/janelia/alignment/filter/CompositeFilter.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public CompositeFilter(final List<Filter> filters) {
+        this.filters = filters;
+    }
+
+    @Override
+    public void init(final Map<String, String> params) {
+        filters = new ArrayList<>(params.size());
+        for (final String serializedFilter : params.values()) {
```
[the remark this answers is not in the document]
minnerbe: You're right, the `values()` are probably in random order (depending on the concrete implementation of the `Map`). Thanks for spotting that!
I kinda prefer the one-entry-per-filter approach since it shows up nicely in the web view. I changed the key to `"filter0": "...", "filter1": "...", ...` and changed `init(...)` so that it retrieves the filters in order. Does this address your concern?

trautmane (approved):
looks good to me

PR #190, opened 2024-05-02
Change order of sanity check and downsample removal

minnerbe:
This is to fix an issue with the append mode of the N5 client: if one of the sanity checks in `setupAppendExportN5` fails, the whole process is aborted. However, the downsampled data was already removed at this point, rendering the existing N5 invalid.

By changing the order, the sanity checks should abort before touching the N5.

There is one thing I'm not sure about: `setupAppendExportN5` updates some metadata of the N5. Is it ok to do this before removing the data, or is the metadata used in the removal process?


trautmane (approved):
I think this should be fine

PR #194, opened 2024-08-16
upgrade to scijava 38.0.1

trautmane:
to get access to latest zstandard compression option


minnerbe (approved):
Thanks so much for cleaning this up, @trautmane ! I think this is ready to be merged.

PR #195, opened 2024-08-28
Refactor grid functionality

minnerbe:
I refactored the Grid class to create a list of `Grid.Block` objects instead of `long[][]` arrays. The latter feels a bit cumbersome and error prone to me (I struggled to remember what the index values meant), so I made it more explicit.

Furthermore, I made `Grid.Block` implement ImgLib2's `Interval`, which makes some uses of the grid blocks more natural, and added clarifying comments to some methods of `Bounds`.


trautmane (approved):
Looks great - thanks for doing this.  I can get rid of the following note on my desk:

Screenshot 2024-09-03 at 3 18 40 PM

PR #197, opened 2024-11-08
Add (interactive) background correction

minnerbe:
This PR adds tools for correcting global shading on large FIBSEM volumes (e.g., [here](https://github.com/JaneliaSciComp/recon_fibsem/issues/89)). These tools are:

* Quadratic and fourth order models that fit the values of a function [(x, y) -> v], where x, y are the pixel coordinates (scaled to [-1, 1] x [-1, 1]) and v is the intensity value.
* A small ImageJ plugin that allows to choose ROIs and fit such models (using only pixel values in the ROIs) interactively.
* A spark client that corrects the intensity of large datasets block-wise.

For supplying model parameters to the spark client, I used an ad hoc json representation, since the model is expected to vary in z. This representation is documented in `BackgroundCorrectionClient`. Feedback about the documentation and the representation itself is very welcome.

One particular problem with doing this at the level of a stack is that we would need an exported mask to not also correct regions where there is no tissue. Is it safe to assume that a mask will always be exported when background correction is done? In this case, I will add support for masks.


minnerbe:
Reassigning @trautmane to have this merged as soon as possible. All of this is new code, so I'm not worried about compatibility issues. Comments about organization into packages, spark parallelization, and usage are highly welcome!

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionClient.java
```diff
@@ -0,0 +1,328 @@
...
+
+        final Map<String, Class<?>> attributes = in.listAttributes(parameters.datasetIn);
+        attributes.forEach((key, clazz) -> {
+            final Object attribute = in.getAttribute(parameters.datasetIn, key, clazz);
+            out.setAttribute(parameters.datasetOut, key, attribute);
+        });
+    }
+
+    private static void processSingleBlock(final Parameters parameters, final ShadingModelProvider modelProvider, final Grid.Block block) {
+        LOG.info("processSingleBlock: block={}", block.gridPosition);
```
trautmane, line 205: Not sure if this is needed, but consider adding something like           
```
 // enable logging on executors and add block context to log messages
 LogUtilities.setupExecutorLog4j(block.gridPosition[0] + ":" + block.gridPosition[1] + ":" + block.gridPosition[2]);
```

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionClient.java
```diff
@@ -0,0 +1,328 @@
...
+            LOG.info("Reading model specs from file: {}", fileName);
+            try (final FileReader reader = new FileReader(fileName)) {
+                return fromJson(JsonParser.parseReader(reader));
+            }
+        }
+
+        public static ShadingModelProvider fromJson(final JsonElement jsonData) {
+
+            final List<ModelSpec> modelSpecs = new ArrayList<>();
+            Collections.addAll(modelSpecs, new Gson().fromJson(jsonData, ModelSpec[].class));
```
trautmane, line 285: To avoid problems with ancient Gson libraries deployed with Spark, I usually use Jackson libraries for JSON stuff (see org.janelia.alignment.json.JsonUtils).  The N5 libraries already depend upon Gson  - so use of it here does not make things worse.  Just wanted to let you know the history/rationale ...

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionTileClient.java
```diff
@@ -0,0 +1,245 @@
...
+            final ShadingModel layerModel,
+            final Parameters parameters,
+            final Bounds bounds
+    ) throws IOException {
+
+        final RenderDataClient renderClient = parameters.webservice.getDataClient();
+        final ResolvedTileSpecCollection rtsc = renderClient.getResolvedTiles(parameters.stack, z);
+
+        if (layerModel == null) {
+            LOG.warn("No model found for z={}", z);
```
trautmane, line 171: If null layer models are expected then this is fine but if it really is an error case I'd throw an exception here.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionTileClient.java
```diff
@@ -0,0 +1,245 @@
...
+        final ShadingModel tileModel = layerModel.copy();
+		try {
+			tileModel.fit(matches);
+		} catch (final NotEnoughDataPointsException | IllDefinedDataPointsException e) {
+			throw new RuntimeException(e);
+		}
+
+        // convert to filter and add to tile spec
+        final ShadingCorrectionFilter filter = new ShadingCorrectionFilter(tileModel);
+        tileSpec.setFilterSpec(FilterSpec.forFilter(filter));
```
trautmane, line 227: Do you need to check if the tile already has a filter spec and then add to a combined filter model here?

trautmane (commented):
Looks great!  I just made a few minor comments/suggestions.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionTileClient.java
trautmane: This will be a common enough case that it might be best to create an `addFilterSpec` method to TileSpec that handles it.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionClient.java
minnerbe: That's actually good to know. I just went with gson because it was easier to make the parsing work. But co-pilot had no problems translating, so I'll use Jackson.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionTileClient.java
minnerbe: They were kind of expected as a way to skip shading correction on certain layers. I made this logic explicit and a little more robust.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/intensityadjust/ShadingCorrectionTileClient.java
minnerbe: Yes, you are right. I vaguely remember writing such a thing in the past, but I can't find it.. Do you maybe remember in which context I might have done this already?

trautmane: I don't - sorry.  I just did a search.  Maybe [this code](https://github.com/saalfeldlab/render/blob/newsolver/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/DistributedIntensityCorrectionSolver.java#L315) is what you were remembering?

minnerbe: Yes, that's it, thanks!

PR #206, opened 2025-03-14
add support for onlyPatchCompletelyUnconnectedTiles option to MFOVMontageMatchPatchClient

trautmane:
add support for onlyPatchCompletelyUnconnectedTiles option to MFOVMontageMatchPatchClient


minnerbe (approved):
Looks good to me. I worked through the logic of finding the completely unconnected tiles and, as far as I can see, it's correct.

However, I found the logic was a bit hard to grasp overall because of all the negations. It would have probably be easier if there was a way of assembling the unconnected tiles directly, i.e., something along the lines
```Java
final Set<String> unconnectedTileIds = new HashSet<>(allTileIds);
for (connectedPair : connectedPairsForPosition) {
  unconnectedTiles.remove(connectedPair.p);
  unconnectedTiles.remove(connectedPair.q);
}
...
for (derivedMatches : derivedMatchesForMFOV) {
  if (unconnectedTiles.contains(p || q))
    completelyUnconnectedTileMatchesList.add(derivedMatches);
}
```
This is very subjective, though, and I doubt that this will worth the effort.

PR #208, opened 2025-03-18
add retry logic to ImageJDefaultLoader to handle occasional failures loading source image data from Google cloud buckets

trautmane:
For my test run of the `DistributedIntensityCorrectionBlockSolverClient` on the `w60_s360_r00_d20_gc_align_b` stack with `212,273` tiles, initial loads failed for `41` tiles but those loads were all successful on the first retry.


minnerbe (approved):
Looks good to me.

As a side note, Google's Python API uses exponential retry times (i.e., `secondsBetweenRetries = 5 << retryNumber` in every call). Not saying that this would be a better strategy in this case, but it could be worth starting with a smaller wait time (e.g., 1s) and then exponentially increase that a couple of times.

PR #212, opened 2025-03-24
newsolver_add_retry_to_imagej_load_process_attempt_b

trautmane:
Add ImageJDefaultLoaderWithTimeout implementation coupled with new IMAGEJ_DEFAULT_W_TIMEOUT LoaderType that wraps each load in a Thread that will time out after 45 seconds.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
+++ b/render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
@@ -28,6 +28,9 @@ public class ImageJDefaultLoader
     /** Shareable instance of this loader. */
     public static final ImageJDefaultLoader INSTANCE = new ImageJDefaultLoader();
 
+    /** The maximum number of retries for each load. */
+    public static int DEFAULT_MAX_RETRIES = 3;
+
     /**
      * @return true (always) because the default ImageJ loader handles 2D sources.
      */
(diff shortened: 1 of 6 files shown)


render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
```diff
@@ -38,21 +41,36 @@ public boolean hasSame3DContext(final ImageLoader otherLoader) {
...
+     *
+     * @return ImagePlus instance for the specified URL.
+     *
+     * @throws IllegalArgumentException
+     *   if the imagePlus cannot be created for any reason.
+     */
+    protected ImagePlus loadImagePlus(final String urlString,
+                                      final int retryNumber,
+                                      final int maxRetries,
+                                      final boolean retryAsNeeded)
```
minnerbe, line 64: It seems like this function is doing two things: loading and retrying. Why does the root class need to be aware of the retrying logic when there's a dedicated subclass for requests that can time out? From our discussions, it seemed to me as if retrying wasn't necessary until now, so I don't see a need for exposing this logic to all subclasses of `ImageJDefaultLoader`.

An alternative could be to push retrying logic down to `ImageJDefaultLoaderWithTimeout` and make `retryNumber` and `maxRetries` fields of that class.

render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
```diff
@@ -117,10 +143,17 @@ private ImagePlus loadWithRetries(final String urlString,
...
+                    LOG.info("loadImagePlus: null imagePlus for {}, will run retry number {} of {} in {} seconds",
+                             urlString, nextRetryNumber, maxRetries, getRetrySleepSeconds(nextRetryNumber));
+                    imagePlus = loadImagePlus(urlString,
+                                              nextRetryNumber,
+                                              maxRetries,
+                                              true);
+                } else {
+                    LOG.info("loadImagePlus: null imagePlus for {} on retry number {} of {}",
+                             urlString, retryNumber, maxRetries);
+                }
```
minnerbe, line 156: This is almost a verbatim duplicate of the code above, is there a way of refactoring this? E.g., the exception could be logged before the snippet and then it's basically the same code.

trautmane:
In case it interests others ... in a test run this morning, 57 out of 104,411 load requests timed-out once and needed to be retried.

render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoaderWithTimeout.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        return imagePlus.getProcessor();
+    }
+
+    private ImagePlus loadImagePlusWithTimedThread(final String urlString,
+                                                   final int retryNumber,
+                                                   final int maxRetries) {
+        ImagePlus imagePlus = null;
+        final ExecutorService executor = Executors.newSingleThreadExecutor();
```
minnerbe, line 47: Creating a new single-threaded executor instance for every load seems a bit wasteful. A quick look in the Goetz book suggests a static `Executors.newCachedThreadPool` (which is unbounded and scales the number of threads as needed) as an alternative.

minnerbe (approved):

minnerbe:
Thanks for the heads-up @trautmane! I will put in the new loader type before the next series of uploads. Is there an easy way for you to patch existing uploads, or should I just rerun my client (without actually uploading) to update the stacks?

render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
trautmane: There are multiple retry scenarios that get triggered by failures in different places.  The ImageJDefaultLoader base class includes the retry scenarios that are "fast/cheap" so that they can just be done by default for most data. The timeout scenario is a separate, relatively expensive one because a thread-pool and thread have to be created for each load request.  That is what motivated the separate ImageJDefaultLoaderWithTimeout implementation.

render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoader.java
trautmane: They are similar, but one includes exception traceback info and the other does not - that's why I have both.

render-app/src/main/java/org/janelia/alignment/loader/ImageJDefaultLoaderWithTimeout.java
trautmane: That is a good point.  I was worried about access to a shared pool getting blocked with large concurrent task sets, but if the unbounded scaling stuff works well then a shared pool would be better.  I'll add this as a TODO for some experimental runs later.

trautmane:
> Is there an easy way for you to patch existing uploads

Yes - I can handle the data you already processed.  We'll just need the new loader type in your stuff for future data.

minnerbe:
Great! I went ahead and added the loader type (see [here for details](https://github.com/JaneliaSciComp/EM_recon_pipeline/pull/36/commits/94149b356b5878b97929c013083ea677c4787554)). Let me know if this is not what you have intended.

PR #215, opened 2025-04-18
fix sfov position issues step 2

trautmane:
Add stitchOnly option to the DistributedAffineBlockSolverClient so that it can be used to solve z layers isolated and in parallel to produce rough aligned (_rxy) stacks.


minnerbe (approved):
Looks good to me!

PR #219, opened 2025-06-05
Feature: variable regularization parameter

minnerbe:
I added support for match number-dependent modification of the alignment model weights. To achieve this, the relative weight of the affine model is multiplied with a factor between 0 and 1, where
- 0 means that there is no affine part (pure regularization, whatever the regularizer model is),
- 1 means that the model is taken as specified (with the full regularizer).

So, in essence, the modification interpolates between "pure regularizer" and "regularized model as given", where the modification factor follows a sigmoidal curve. 

The behavior of this can be specified by `--weightModificationDeciles LxH`, where
- L is the lowest decile: tiles with this number of matches get a factor of 0.1,
- H is the highest decile: tiles with this number of matches get a factor of 0.9.

If this option is not given, there is no weight modification, i.e., the alignment behaves as before.
A stack where this weighting with parameter `50x500` was used [can be found here](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B78393.125%2C102653.3515625%2C1%5D%2C%22crossSectionScale%22:5.712111369106358%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafers_60_61/w60_serial_360_to_369/w60_s360_r00_d20_gc_dw_mi_test_2%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22w60_s360_r00_d20_gc_dw_mi_test_2%22%7D%5D%2C%22selectedLayer%22:%7B%22layer%22:%22w60_s360_r00_d20_gc_dw_mi_test_2%22%7D%2C%22layout%22:%22xy%22%7D).

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/affine/AffineAlignBlockWorker.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/affine/AffineAlignBlockWorker.java
@@ -886,7 +886,7 @@ public class AffineAlignBlockWorker<M extends Model<M> & Affine2D<M>, S extends 
 		if (LOG.isInfoEnabled()) {
 			final DoubleSummaryStatistics errors = SolveTools.computeErrors(tileConfig.getTiles());
 			LOG.info("solve: block {}, optimizing {} tiles with preAlign {}, error stats before optimization are {}",
-					 blockData, solveItem.groupedTileToTiles().keySet().size(), preAlign, errors);
+					 blockData, solveItem.groupedTileToTiles().size(), preAlign, errors);
 		}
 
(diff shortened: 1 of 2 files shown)


trautmane (approved):
looks good to me

PR #228, opened 2025-10-31
black line at tile border render problem

trautmane:
Tobias' LineMapper improvement committed in December 2024 and January 2025 might have introduced a problem where we get black lines at tile edges under certain circumstances.  My initial commit here is just a test case that quickly demonstrates the problem with some multi-SEM data I was working with earlier this week.  My hope is that Tobias will be able to take a look and figure out what actually needs to be fixed.

Here is what the problem looks like:

black-line-problem


trautmane:
Fixed by https://github.com/saalfeldlab/render/pull/247

PR #233, opened 2026-06-01
Feature/better normalization

minnerbe:
This PR adds per-layer histogram normalization for multisem stacks, mimicking Michal's workflow. It runs at the individual-tile level (before stack assembly), making the whole process very efficient in terms of memory (no stack copy is needed, only filter metadata is written). Additions:

  - MichalLayerNorm: histogram-matches each z-layer to a reference layer and applies the resulting 256-entry LUT to every tile. Above-threshold pixels are mapped by CDF percentile lookup; below-threshold pixels are filled with a monotone cubic (PCHIP) interpolation from the lowest matched percentile to (0,0), biased toward leaving sub-threshold pixels unchanged.
  - LutFilter: new 8-bit Filter that applies a 256-entry LUT (clipped to [0,255]) via ImageJ's applyTable, serialized as a comma-separated parameter string.
 
One thing to note: The serialization takes about 1kb, which might add up to a significant amount since it's stored for each tile and it's the same for every tile in a given layer. I have no good intuition about how much the render database can reduce this redundancy automatically, though.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/filter/LutFilter.java
+++ b/render-app/src/main/java/org/janelia/alignment/filter/LutFilter.java
@@ -1,0 +1,89 @@
+package org.janelia.alignment.filter;
+
+import ij.process.ByteProcessor;
+import ij.process.ImageProcessor;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


minnerbe:
This PR was abandoned in favor of other approaches. Closing it.

PR #235, opened 2026-07-16
first draft of layer-as-tile code

trautmane:
This code is intended to "roughly-3D-align" the 2D-stitch-only solve results to reduce the amount of work needed to run the results through SOFIMA.  It copies/reuses many of the components developed for our MFOV-as-tile process.

The process works with and creates a set of stacks like this:
```
w61_s109_r00_gc_par_crc_aso :             starting 2D-stitch-only SFOV tile stack
w61_s109_r00_gc_par_crc_aso_lat :         layer-as-tile stack with dynamically rendered layer images
w61_s109_r00_gc_par_crc_aso_latr :        layer-as-tile stack with persisted layer images
w61_s109_r00_gc_par_crc_aso_latr_align :  aligned layer-as-tile stack with persisted layer images
w61_s109_r00_gc_par_crc_aso_3d :          rough 3D aligned SFOV tile stack
```

The process also creates a layer-as-tile match collection that is used for the layer-as-tile solve.

Our hope is that the `..._aso_3d` result stacks are good enough that we can skip cross-layer SFOV match data generation and that the SOFIMA setup work will be reduced/simplified.

A few final notes ...
- We originally thought the layer-as-tile stacks might be useful for storing layer-based histogram intensity correction information, but that process has changed.  We currently do not need to store histogram data.  
- The layer-as-tile stacks might be useful if we want to orient the slabs similarly before we try to connect them into a single volume.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/multisem/Layer.java
+++ b/render-app/src/main/java/org/janelia/alignment/multisem/Layer.java
@@ -59,8 +59,8 @@ public class Layer {
 
         final int numberOfTilesInZLayer = renderParameters.numberOfTileSpecs();
 
-        final int scaledImageWidth = (int) Math.floor(renderParameters.width * renderScale);
-        final int scaledImageHeight = (int) Math.floor(renderParameters.height * renderScale);
+        final double scaledImageWidth = Math.floor(renderParameters.width * renderScale);
+        final double scaledImageHeight = Math.floor(renderParameters.height * renderScale);
         final double x = renderParameters.x * renderScale;
         final double y = renderParameters.y * renderScale;
 
@@ -68,8 +68,8 @@ public class Layer {
 
         tileSpec.setTileId(toLayerAsTileName(stackId.getStack(), z.intValue()));
         tileSpec.setZ(z);
-        tileSpec.setWidth((double) scaledImageWidth);
-        tileSpec.setHeight((double) scaledImageHeight);
+        tileSpec.setWidth(scaledImageWidth);
+        tileSpec.setHeight(scaledImageHeight);
 
         final LayoutData layoutData = new LayoutData(z.toString(), null, null,
                                                      0, 0, x, y, null);
--- a/render-app/src/main/java/org/janelia/alignment/multisem/Layer.java
+++ b/render-app/src/main/java/org/janelia/alignment/multisem/Layer.java
@@ -1,0 +1,117 @@
+package org.janelia.alignment.multisem;
+
+import java.util.Collections;
+
+import mpicbg.trakem2.transform.TranslationModel2D;
+
+import org.janelia.alignment.ImageAndMask;
+import org.janelia.alignment.RenderParameters;
+import org.janelia.alignment.loader.ImageLoader;
+import org.janelia.alignment.spec.ChannelSpec;
+import org.janelia.alignment.spec.LayoutData;
+import org.janelia.alignment.spec.LeafTransformSpec;
+import org.janelia.alignment.spec.TileSpec;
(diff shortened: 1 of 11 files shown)


render-app/src/main/java/org/janelia/alignment/multisem/Layer.java
```diff
@@ -0,0 +1,117 @@
...
+                                         final Double z,
+                                         final double renderScale) {
+
+        final String renderParametersUrl = buildRenderParametersUrl(baseDataUrl, stackId, z, renderScale);
+        final RenderParameters renderParameters = RenderParameters.loadFromUrl(renderParametersUrl);
+
+        final int numberOfTilesInZLayer = renderParameters.numberOfTileSpecs();
+
+        final int scaledImageWidth = (int) Math.floor(renderParameters.width * renderScale);
+        final int scaledImageHeight = (int) Math.floor(renderParameters.height * renderScale);
```
minnerbe: These variables could be doubles. The only usage is a few lines below, where they're explicitly cast to double again.

render-ws-java-client/src/main/java/org/janelia/render/client/multisem/MFOVAsTileStackClient.java
minnerbe, line 1: I have very minor and abstract concerns about the extensibility and robustness of the x-as-tile abstraction that I find hard to articulate. @trautmane if you have the same feeling, I'm happy to discuss this offline. Otherwise, I'd view this as an instance of the WET-principle and revisit this abstraction when (and if) we need to extend this a third time.

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
```diff
@@ -0,0 +1,203 @@
...
+        setup.stitching.maxIterations = 5000;
+        setup.stitching.maxPlateauWidth = 1000;
+        setup.stitching.minInliers = 25;
+
+        setup.blockOptimizer.lambdasRigid = List.of(1.0, 1.0, 0.9, 0.3, 0.01);
+        // NOTE: lambda's translation means how much do you want to regularize with a translation model, thus 1.0 means 100% translation
+        setup.blockOptimizer.lambdasTranslation = MFOVAsTileParameters.SolveType.AFFINE.getLambdasTranslation();
+        setup.blockOptimizer.lambdasRegularization = List.of(0.0, 0.0, 0.0, 0.0, 0.0);
+
+        setup.blockOptimizer.iterations = List.of(1000, 1000, 500, 250, 250);
```
minnerbe, line 163: Are the first two rounds of optimization exactly the same (1000 iterations of 100% rigid with plateau width 250)? In this case, we could simply concatenate them (i.e., 2000 iterations of 100% rigid with plateau width 250).

minnerbe (approved):
Thanks, @trautmane! Except for some small comments (and an optional offline conversation), this looks good to me.

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
trautmane: The lambdasTranslation go from 1.0 in round 1 to 0.0 in round 2, so I don't think we should concatenate the first two rounds.  I could be misunderstanding things though.  Let me know if I am wrong.

minnerbe: Thanks for the clarification! I didn't see the translation parameters, so if they do vary from round 1 to round 2, my comment is obsolete.

PR #236, opened 2026-07-21
Optimize intensity matching

minnerbe:
This PR slightly shifts the level of caching from raw tiles to downscaled and filtered or even fully rendered tiles. The first one is a zero-tradeoff improvement: instead of the full-scale tile + mask, the downsampled versions are stored, reducing per-tile effort and cache pressure. Therefore, it is the new default. The second one trades memory for computational resources: the full rendered bounding box is stored, making the first render somewhat expensive, and the following ones essentially free. It should therefore be used cautiously (especially for ic3d processes, where the full tile needs to be rendered anyway).

@trautmane This introduces a hand-rolled cache that was modeled after `ImageProcessorCache`. If you think it's worth doing that, I'm happy to try to consolidate all the different caching mechanisms. Also, imglib2 images are used instead of ImageJ image processors to enable zero-copy views for the fully-cached implementation.

## Previous state

`IntensityMatcher` rendered both tiles of every overlapping pair from scratch, on every call, via `Render.render(...)` into ImageJ `FloatProcessor`/`ColorProcessor` rasters sized to just the overlap box. The only cache in the path was a shared `ImageProcessorCache` of raw, full-resolution source images (pre-filter, pre-downsample, pre-mesh) — so filtering, downsampling, and mesh-mapping were all redone for every pair a tile appears in, even though a tile typically overlaps many same-layer and cross-layer neighbors.

## Two options now

Rendering is factored into `TileRenderer`, split into two independent halves:

- **`loadSource`** — pair-independent: load the full-res source, apply the tile's filter, downsample to its mipmap level → `DownsampledSource`.
- **`renderBox`** — pair-dependent: transform the downsampled source into the requested world-space box (using a mesh transform) → `RenderedTile` (image / weight / coefficient-label rasters).

Two strategies decide what to cache between calls, selected by the new `--cacheRenderedTiles` flag:

| | `TightBoxTileRenderer` (default) | `CachedTileRenderer` (`--cacheRenderedTiles`) |
|---|---|---|
| Caches | `DownsampledSource` per tile | full-tile `RenderedTile` (whole bounding box) per tile |
| Per-pair cost | re-mesh the box every time | zero-copy crop (`Views.interval`) of the cached render |
| Tradeoff | less memory, more compute | more memory, less compute |

Both share a small generic LRU, `TileCache<T>`, keyed by tile spec and weight-bounded in kilobytes by the existing `--maxPixelCacheGb` (adapted from `ImageProcessorCache`'s eviction logic). As before, same-layer and cross-layer matching share one renderer/cache when their render scales are equal.

## New / modified classes

**New:**
- `TileRenderer` — abstract base defining the `loadSource` / `renderBox` split and the shared mesh-mapping logic.
- `TightBoxTileRenderer` — caches only the downsampled source; re-meshes each overlap box.
- `CachedTileRenderer` — caches the full per-tile render; serves overlaps as zero-copy crops.
- `TileCache<T>` — generic per-tile-id, kilobyte-bounded LRU cache used by both renderer strategies.
- `DownsampledSource` — the cacheable pair-independent render result (filtered/downsampled image, mask, mipmap level).
- `RenderedTile` — the image/weight/coefficient rasters as imglib2 views, replacing the previous raw ImageJ processors.

**Modified:**
- `IntensityMatcher` — no longer renders directly; picks a same-/cross-layer `TileRenderer` and iterates the resulting imglib2 rasters in lockstep instead of indexing ImageJ processors by hand.
- `AffineIntensityCorrectionBlockWorker` — replaced the single shared `ImageProcessorCache` with a kilobyte budget passed down to `IntensityMatcher`, which now owns its own per-scale `TileCache` instances.
- `FIBSEMIntensityCorrectionParameters` — carries the new `cacheRenderedTiles` flag through to the matcher.
- `AlgorithmicIntensityAdjustParameters` — adds the `--cacheRenderedTiles` CLI flag.

## Validation and benchmarking

The three methods give exactly the same results, see results obtained with the:
- [old process](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B47335.8828125%2C47562.84765625%2C2.5%5D%2C%22crossSectionScale%22:4.086090815809243%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafers_60_61/w61_serial_100_to_109/w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%2C%22layout%22:%22xy%22%7D)
- [new default](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B47335.5%2C47562.5%2C2.5%5D%2C%22crossSectionScale%22:4.086090815809243%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafers_60_61/w61_serial_100_to_109/w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm2%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%2C%22layout%22:%22xy%22%7D)
- [new process using the `--cacheRenderedTiles` flag enabled](http://renderer.int.janelia.org:8080/ng/#!%7B%22dimensions%22:%7B%22x%22:%5B8e-9%2C%22m%22%5D%2C%22y%22:%5B8e-9%2C%22m%22%5D%2C%22z%22:%5B8e-9%2C%22m%22%5D%7D%2C%22position%22:%5B47335.5%2C47562.5%2C2.5%5D%2C%22crossSectionScale%22:4.086090815809243%2C%22projectionScale%22:32768%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22render://http://renderer.int.janelia.org:8080/hess_wafers_60_61/w61_serial_100_to_109/w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm3%22%2C%22subsources%22:%7B%22default%22:true%2C%22bounds%22:true%7D%2C%22enableDefaultSubsources%22:false%7D%2C%22tab%22:%22source%22%2C%22name%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22w61_s109_r00_gc_par_align_thomas_inv2_ic3d_bm1%22%7D%2C%22layout%22:%22xy%22%7D)

Runtimes with the processes can be seen below, where 10 layers of the w61_s109 multi-sem stack were corrected with ic3d. Matching times were recorded, clustered into two clusters (roughly cached and uncached), and the median of both clusters is shown, as well as overall runtime and cache miss overhead. The first image shows the situation with an overly large cache (100GB), the second with a somewhat tight cache (1GB). It looks like 10GB should be sufficient for multi-sem stacks for all three methods. In the case where the cache is sufficiently large, the new methods speed up the process by ~3x and ~5x, respectively.

image
image

FYI @StephanPreibisch


render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/RenderedTile.java
trautmane, line 1: If this is or can be just a region of a tile as the javadoc states, I suggest that you include "Region" in the class name.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/TileCache.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.util.concurrent.UncheckedExecutionException;
+
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+import java.util.function.ToIntFunction;
+
+import org.janelia.alignment.spec.TileSpec;
+
+/**
+ * A small weight-bounded, least-recently-used cache of per-tile values, keyed by tile id and backed
```
trautmane, line 15: Very minor comment suggestion: remove "least-recently-used" here.  It initially made me think the cache was holding least-recently-used values as opposed to evicting them.  You mention the eviction policy a few lines down.

trautmane (approved):
Looks good to me!  My first instinct is to consolidate the caching additions with the other implementations but it is not that important.  Let's leave this as-is for now and come back to consolidate later if we think consolidation (or improvement of the older cache stuff) is worth the effort.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/RenderedTile.java
minnerbe: ~~It _can_ be, but it's never used this way. I'm currently leaning toward just changing the javadoc. Thanks!~~

Looking at it more closely, it is used as a region most of the time. I renamed it.

render-ws-java-client/src/main/java/org/janelia/render/client/newsolver/solvers/intensity/TileCache.java
minnerbe: Huh, interesting. This is just the term "LRU cache" (that I think is fairly standard) spelled out, but I agree that in this form my first thought would be the same as yours. I'll just rephrase it at "LRU cache" since that is the standard term.

minnerbe:
Thanks for your comments, @trautmane! In response to them, I renamed `RenderedTile` to `RenderedRegion` and updated `TileCache`'s docstring (see [here](https://github.com/saalfeldlab/render/commit/9fd36e102caee2f04cdecb8d9c4138f2a06a19aa)).

On a related note, I also deprecated the "old" intensity matching code that we haven't used in a while (see [here](https://github.com/saalfeldlab/render/commit/50863ecf45f870da040c45a52366671805c8e9cb)).

I agree that we should consolidate the caching implementations at some point. It was not immediately clear to me how to do that, since `ImageProcessorCache` is a bit more powerful than what I implemented here. Happy to talk about this when there is a need/desire to touch this again.

PR #237, opened 2026-07-21
Use calibration data to precondition intensity correction

minnerbe:
This PR uses some of the calibration data acquired and processed by the Hess lab to precondition the intensity correction for multi-sem data. This improves the outcome of subsequent intensity correction steps a lot and also helps to equalize intensity between layers.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/filter/LinearIntensityMap8BitFilter.java
+++ b/render-app/src/main/java/org/janelia/alignment/filter/LinearIntensityMap8BitFilter.java
@@ -43,9 +43,19 @@ public class LinearIntensityMap8BitFilter
     public void after(final LinearIntensityMap8BitFilter otherFilter) {
         final double[][] otherCoefficients = otherFilter.getCoefficients();
 
+        // A 1x1 (single region) filter is spatially constant, so it is broadcast across all regions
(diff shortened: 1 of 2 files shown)


minnerbe:
Hi @trautmane, I noticed that you just pushed the last three commits. Initially, I intentionally left these out of this PR for two reasons:
1. They are off-topic for the correction based on calibration data.
2. They need some consolidation. I found that only `--skipMatchesAboveDifference 0` yields reasonable results, which is why I would simplify the whole process a little bit and just introduce a `--onlyTetherCrossLayers` flag that has the same effect without the (in hindsight) unnecessary choice of cut-off parameter.

I don't exactly remember what I did to limit the scope of this PR, but I assume it involved some history rewriting and force-pushing, which I should have communicated clearly. Sorry about that!

How do you suggest we proceed?

trautmane:
Oh sorry @minnerbe - I was trying to get started on adding support for use of the intensity correction within a spark alignment pipeline and on adding Thomas' tile re-ordering.  My local copy of this branch must have had your earlier changes.  Since you know what belongs and what does not ... can you remove whatever needs to be removed again and merge the remaining changes into the newsolver branch?  I will then create a new branch just for my changes.  Sorry again for complicating this process.

PR #243, opened 2026-08-10
feature/thomas-intensity-changes-b

trautmane:
This pull request involves changing tile identifiers to improve render order, adding beam intensity correction as a spark pipeline step, and persisting/reusing within-MFOV match data during beam intensity correction and MFOV-as-tile processing.

### DIFF
--- a/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
+++ b/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
@@ -201,13 +201,13 @@ public class ResolvedTileSpecsWithMatchPairs
             } else {
 
                 final StringBuilder sb = new StringBuilder();
-                if (pTileSpec == null) {
-                    if (qTileSpec == null) {
-                        sb.append("s ").append(pair.getpId()).append(" and ").append(pair.getqId());
+                if (qTileSpec == null) {
+                    if (pTileSpec == null) {
+                        sb.append("s ").append(pair.getpId()).append(" and ");
                     }
-                    sb.append(" ").append(pair.getpId());
-                } else {
                     sb.append(" ").append(pair.getqId());
+                } else {
+                    sb.append(" ").append(pair.getpId());
                 }
                 throw new IllegalStateException("missing tile" + sb + " for match pair " + pair.toKeyString());
 
--- a/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
+++ b/render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
@@ -12,6 +12,7 @@ import java.util.stream.Collectors;
 
 import org.janelia.alignment.json.JsonUtils;
 import org.janelia.alignment.match.CanvasMatches;
+import org.janelia.alignment.multisem.MultiSemUtilities;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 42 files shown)


render-app/src/main/java/org/janelia/alignment/spec/ResolvedTileSpecsWithMatchPairs.java
```diff
@@ -159,6 +159,63 @@ public void normalize(final Integer maxZDistance,
...
+
+                final StringBuilder sb = new StringBuilder();
+                if (pTileSpec == null) {
+                    if (qTileSpec == null) {
+                        sb.append("s ").append(pair.getpId()).append(" and ").append(pair.getqId());
+                    }
+                    sb.append(" ").append(pair.getpId());
+                } else {
+                    sb.append(" ").append(pair.getqId());
+                }
```
minnerbe:
```suggestion
                if (qTileSpec == null) {
                    if (pTileSpec == null) {
                        sb.append("s ").append(pair.getpId()).append(" and ");
                    }
                    sb.append(" ").append(pair.getqId());
                } else {
                    sb.append(" ").append(pair.getpId());
                }
```

minnerbe: In the current formulation, `pId is printed twice if both tile specs are `null`. This change should print `p`, `q`, or `s p and q` correctly, depending on the situation.

render-app/src/main/java/org/janelia/alignment/spec/TileBoundsRTree.java
```diff
@@ -334,6 +338,8 @@ public static Set<OrderedCanvasIdPair> getDistinctPairs(final TileBounds fromTil
 
                         if (excludeSameMfovNeighbors && (pMfov.equals(MultiSemUtilities.getMagcMfovForTileId(qTileId)))) {
                             continue;
+                        } else if (excludeDifferentMfovNeighbors && (! pMfov.equals(MultiSemUtilities.getMagcMfovForTileId(qTileId)))) {
```
minnerbe, line 342: For readability, consider declaring `final boolean isSameMfov = pMfov.equals(MultiSemUtilities.getMagcMfovForTileId(qTileId))`.

render-ws-java-client/src/main/java/org/janelia/render/client/multisem/BeamCorrectionClient.java
```diff
@@ -324,54 +270,82 @@ private static FilterSpec buildHomogenizationFilterSpec(final double gain,
...
+	 * Reads a JSON metadata resource (e.g. {@code .zarray} or {@code .zattrs}) for a dataset,
+	 * falling back to the consolidated {@code .zmetadata} at the container root.
+	 * <p>
+	 * The resource is read with the container's {@code KeyValueAccess} instead of {@link java.nio.file.Files}
+	 * so that metadata can be read from a local file system path or from a cloud URI
+	 * (e.g. {@code gs://bucket/xlog_wafer_61.zarr}).
+	 *
+	 * @return the parsed resource or null if it is not present in the container.
+	 */
+	private static JsonObject readDatasetJson(final ZarrKeyValueReader reader,
```
minnerbe, line 305: Just a general comment on this file: I've had to read the metadata as plain json instead of going through N5's key-value reader because it cannot deal with some of the values that are written by python. This is fixed in a newer N5 version, which is pulled in in #239. This file is also significantly simplified there.

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/tile/TileIdHackClient.java
```diff
@@ -0,0 +1,163 @@
...
+import org.janelia.render.client.spark.pipeline.AlignmentPipelineStep;
+import org.janelia.render.client.spark.pipeline.AlignmentPipelineStepId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Spark client for updating the tileId for all tiles in the stack
+ * to use the new wafer 60 and 61 render order.
+ */
+public class TileIdHackClient
```
minnerbe, line 36: I think I'd prefer something like `TileIdUpdateClient` or `ChangeTileIdClient`. "Hack" sounds like we're doing something wrong.. ;-)

minnerbe (approved):
Thanks for putting this together, @trautmane! I have skimmed the code and only dug into a few targeted spots. All looks good to me.  As always, my review comments are mostly just suggestions.

render-app/src/main/java/org/janelia/alignment/spec/TileBoundsRTree.java
trautmane: I agree adding isSameMfov would improve readability, but because this spot in the code is heavily used for FIB-SEM data sets I prefer to save time (for the FIBSEM data sets) by not doing the multi-SEM string conversion and comparison unless the parameters indicate this is a multi-SEM stack.

render-ws-java-client/src/main/java/org/janelia/render/client/multisem/BeamCorrectionClient.java
trautmane: Good to know -thanks!

render-app/src/main/java/org/janelia/alignment/spec/TileBoundsRTree.java
trautmane: I added a comment to explain this in the code

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/tile/TileIdHackClient.java
trautmane: Yeah - I agree that it sounds like we are doing something wrong.  My initial naming rationale was consistency with the MaskHackClient and my preference for thinking of tile ids as read only things (meaning in the future we should set the tile ids "properly" at the start rather than updating them later).  I'm going to leave this as is for now out of laziness, but I agree with you.

PR #246, opened 2026-08-21
added support for using peak scan data in tile removal

trautmane:


minnerbe (approved):
Thanks, @trautmane! I skimmed all of it, but paid special attention to how the number is handled (z value instead of scan), which seems to be correct.

PR #249, opened 2026-08-29
improve-layer-as-tile-20260828

trautmane:
This contains various improvements to the layer-as-tile process.

### DIFF
--- a/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
+++ b/render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
@@ -11,6 +11,7 @@ import org.janelia.alignment.match.ModelType;
 import org.janelia.alignment.match.parameters.FeatureRenderClipParameters;
 import org.janelia.alignment.match.parameters.MatchRunParameters;
 import org.janelia.alignment.match.parameters.MatchStageParameters;
+import org.janelia.alignment.spec.Bounds;
 import org.janelia.alignment.spec.stack.StackId;
 import org.janelia.render.client.newsolver.blocksolveparameters.FIBSEMAlignmentParameters;
 import org.janelia.render.client.newsolver.setup.AffineBlockSolverSetup;
@@ -125,10 +126,20 @@ public class LayerAsTileParameters
         return this.renderedLayerRunTimestamp;
     }
 
-    public List<MatchRunParameters> buildLayerMatchRunList() {
+    public List<MatchRunParameters> buildLayerMatchRunList(final Bounds renderedLayerStackBounds) {
         final List<MatchRunParameters> layerMatchRunList = new ArrayList<>();
-        layerMatchRunList.add(buildCrossMatchRunParameters());
+        layerMatchRunList.add(buildCrossMatchRunParameters(renderedLayerStackBounds));
         return layerMatchRunList;
+    }
+
+    /**
+     * @return the minimum number of match inliers to require for layers with the specified bounds.
+     *         Larger layers need more inliers, so the count is derived from the layer area
+     *         instead of being hard-coded for layers of any size.
+     */
+    public static int deriveMatchMinNumInliers(final Bounds renderedLayerStackBounds) {
+        final double layerPixelCount = renderedLayerStackBounds.getDeltaX() * renderedLayerStackBounds.getDeltaY();
+        return (int) Math.round(layerPixelCount / PIXELS_PER_MIN_INLIER);
     }
 
     public AffineBlockSolverSetup buildLayerAffineBlockSolverSetup() {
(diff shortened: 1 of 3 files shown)


render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/LayerAsTileClient.java
```diff
@@ -297,49 +306,125 @@ private static void generateLayerAsTileMatches(final JavaSparkContext sparkConte
...
+                                                                               matchCollectionId,
+                                                                               renderDataClient,
+                                                                               jcccp.tileCluster);
+            } catch (final Exception e) {
+                throw new IOException(e);
+            }
+
+            final String countErrorString = summary.buildCountErrorString(1,
+                                                                          0,
+                                                                          0);
```
minnerbe, line 417:
```suggestion
            final String countErrorString = summary.buildCountErrorString(1, 0, 0);
```

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileStackLists.java
```diff
@@ -77,6 +85,53 @@ public  List<StackWithZValues> getRenderedLayerStacksWithAllZ() {
         return renderedLayerStacksWithAllZ;
     }
 
+    /**
+     * Loads the bounds for all rendered layer stacks.
+     *
+     * @throws IOException
+     *   if the bounds for any rendered layer stack cannot be retrieved.
+     */
+    public void loadRenderedLayerStackBounds()
```
minnerbe: I think this method adds some cognitive overhead for users of the class. They must call the two methods in order, failure to do so results in a runtime error. Would it be possible to do the following instead?
- Have `loadRenderedLayerStackBounds` return a `Map<String, Bounds>`
- Lazily assemble the map when requested, but return a cached version if already loaded (i.e., guard behind a `if (map == null) {map = load();} return map;`-type method)

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
```diff
@@ -200,4 +215,16 @@ private static MatchRunParameters buildCrossMatchRunParameters() {
...
+     * Number of layer pixels for each required match inlier.
+     * Layers with 12 million pixels typically need a minimum of 100 inliers.
+     */
+    private static final double PIXELS_PER_MIN_INLIER = 120_000.0;
+
+    /**
+     * Derived minimum inlier counts are rounded to a multiple of this factor so that
+     * stacks with similar layer sizes can be batched together for match generation.
+     */
+    private static final int MIN_INLIER_ROUNDING_FACTOR = 25;
```
minnerbe, line 229: Maybe I don't understand how the actual match derivation is done well enough, but why does that need to be batched by minimum number of inliers?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/LayerAsTileClient.java
```diff
@@ -297,49 +306,125 @@ private static void generateLayerAsTileMatches(final JavaSparkContext sparkConte
...
 
                     matchClient.generatePairsAndMatchesForRunList(sparkContext,
                                                                   multiProject,
-                                                                  listOfRenderedLayerStackLayersInProject,
+                                                                  batchLayers,
                                                                   layerMatchRunList);
                 }
 
+                validateEachStackIsConnectedWithOneMatchCluster(baseDataUrl,
+                                                                projectStacks);
```
minnerbe, line 370:
```suggestion
                validateEachStackIsConnectedWithOneMatchCluster(baseDataUrl, projectStacks);
```

minnerbe (commented):
Looks good to me, in principle. I think some of the loading logic for the layers-as-tiles stacks may be a bit improved and I don't fully understand the reason for batching match derivation by minimum number of inliers. Happy to discuss this offline if it helps.

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileStackLists.java
trautmane: Good point - I simplified this with a commit below.

render-ws-java-client/src/main/java/org/janelia/render/client/parameter/LayerAsTileParameters.java
trautmane: The MultiStagePointMatchClient will parallelize match generation for pairs in different stacks, but only if the match parameters are the same.  Originally the same match parameters were used for all stacks so it worked fine.  But, the introduction of different minInliers for different stacks (now needed to support small region slabs) created slightly different minInliers for almost all stacks.  Many stacks have similar but not identical sizes.  This rounding allows the similar sized stacks to be grouped together and processed - improving concurrency.

PR #253, opened 2026-09-13
import_sofima_20260912

trautmane:
Added spark client for importing SOFIMA results and cleaned-up AlignmentPipelineStep interface.


trautmane:
@minnerbe - I went ahead and merged this, but please let me know if you think anything should be changed.  Thanks!

trautmane:
I needed to add this to get things working:
https://github.com/saalfeldlab/render/commit/c89b17183cd6953d04e8fd50ed047c9ff4c54453

trautmane:
Also - the log from my first run contains many applyInPlace inversion warnings like these:

applyInPlace-inversion-convergence-warning

Please let me know if this is expected or if it identifies a problem.
Thanks!

minnerbe:
Thanks, @trautmane ! The fix looks good to me. I'm a bit concerned about the warnings, especially the one that has a 33px residual. I think the screenshot has all information I need to investigate this.

minnerbe:
I investigated the issue and it turned out that the convergence properties of the inversion of the displacement field weren't great. I implemented a Newton method, which should converge much more robustly (getting rid of the warnings above, which indicated that the inversion was wrong). This has been pushed [directly to newsolver](https://github.com/saalfeldlab/render/commit/d3422e17c28726a38fdeeb8b1bdcc9246591a981).

PR #254, opened 2026-09-23
fix-creep-match-issue-20260917

trautmane:
Moved creep correction after all match derivation in the pipeline process to fix/work around problems introduced when within-MFOV matches were generated before creep correction and cross-MFOV matches were generated after creep correction.  This change meant that the new java 21 and scijava code was run for the first time on the rough alignment pipeline which exposed a couple of new problems when running Google Dataproc batch jobs.

The following commit fixed the core issue:
https://github.com/saalfeldlab/render/commit/366e017956d7ec71cf6e2cc7b34b37341de10da7

This pull request also contains improvements to data management scripts used in the Google VMs.

Finally, the mfov-as-tile and layer-tile tileIds were changed to include scan names before the z values in the tileIds to make it easier to debug problems later.


trautmane:
I went ahead and merged this.  If you notice any problems, please post them here and I'll address them in another fix branch.  Thanks!

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/StackRenameClient.java
minnerbe: The comments in this file make it very clear that there is no spark parallelization here. What is the reason for formulating this as a spark client and putting it into render-ws-spark-client instead of render-ws-java-client?

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/StackRenameClient.java
```diff
@@ -0,0 +1,160 @@
...
+        final Set<StackId> existingStackIdSet = new HashSet<>(existingStackIds);
+        final Map<String, Set<String>> projectToTargetNames = new HashMap<>();
+        for (int i = 0; i < targetStackIds.size(); i++) {
+            final StackId targetStackId = targetStackIds.get(i);
+            if (existingStackIdSet.contains(targetStackId)) {
+                throw new IllegalArgumentException("cannot rename " + sourceStackIds.get(i).getStack() + " to " +
+                                                   targetStackId.getStack() + " because a stack with that name " +
+                                                   "already exists in project " + targetStackId.getProject() +
+                                                   " for owner " + owner);
+            }
```
minnerbe: This will blow up on the first existing stack, hiding potential other collisions. A way of making this check as informative as possible would be something like
```Java
existingStackIdSet.retainAll(targetStackIds);
if (!existingStackIdSet.isEmpty()) { // Report all collisions at once }
```

render-ws-spark-client/src/main/java/org/janelia/render/client/spark/multisem/StackRenameClient.java
```diff
@@ -0,0 +1,160 @@
...
+            final Set<String> targetNamesForProject =
+                    projectToTargetNames.computeIfAbsent(targetStackId.getProject(), p -> new HashSet<>());
+            if (! targetNamesForProject.add(targetStackId.getStack())) {
+                throw new IllegalArgumentException("cannot rename more than one stack to " +
+                                                   targetStackId.getStack() + " in project " +
+                                                   targetStackId.getProject() + " for owner " + owner);
+            }
```
minnerbe: Also this only reports the first problem, although reporting all duplicates is a bit more involved here.

minnerbe (approved):
Looks good to me. I only reviewed the java files, since my limited bash knowledge is unlikely to surface any non-trivial problem ;) I left a few optional comments in the code.

